PR #4, opened 2019-04-16
Add original image field to base bundle

glyn:
See https://github.com/deislabs/cnab-spec/issues/134


bundle/bundle.go
```diff
@@ -67,12 +67,13 @@ type LocationRef struct {
...
 type BaseImage struct {
-	ImageType string         `json:"imageType" mapstructure:"imageType"`
-	Image     string         `json:"image" mapstructure:"image"`
-	Digest    string         `json:"digest,omitempty" mapstructure:"digest"`
-	Size      uint64         `json:"size,omitempty" mapstructure:"size"`
-	Platform  *ImagePlatform `json:"platform,omitempty" mapstructure:"platform"`
-	MediaType string         `json:"mediaType,omitempty" mapstructure:"mediaType"`
+	ImageType     string         `json:"imageType" mapstructure:"imageType"`
+	Image         string         `json:"image" mapstructure:"image"`
+	OriginalImage string         `json:"originalImage" mapstructure:"originalImage"`
```
silvin-lubecki: As far as I can remember `originalImage` is `OPTIONAL`, so I think it should be annotated with `omitempty`:
```go
OriginalImage string         `json:"originalImage,omitempty" mapstructure:"originalImage"`
```

glyn: Correct. Fixed (with a test).

PR #7, opened 2019-05-09
Add initial Brigade pipeline

radu-matei:


radu-matei:
I just LOVE how fast our Brigade pipeline executes 💛

PR #8, opened 2019-05-09
Add more types from Duffle

radu-matei:
As more projects start implementing the CNAB specification in Go, we see them trying to take a direct dependency on Duffle.

This means that any dependency Duffle takes has the potential to become a transitive dependency for the project that imports Duffle - and while tree shaking *should* take care of any unused packages, this doesn't always happen - leaving projects that import Duffle in a state where they have to very closely follow their imported dependencies - even though the packages they are trying to use are mostly self-contained.

Thus, this PR introduces the following packages from Duffle:

- action
- bundle
- claim
- credential
- driver
- utils (used for the crud package used by the claim store - note that it doesn't introduce additional external dependencies, and it is quite useful)

This PR introduces the external dependencies:

```
(1/6) Wrote github.com/pmezard/go-difflib@v1.0.0 - used by testify
(2/6) Wrote gopkg.in/yaml.v2@v2.2.2 - direct dependency
(3/6) Wrote github.com/davecgh/go-spew@v1.1.1 - used by testify
(4/6) Wrote github.com/docker/go@v1.5.1-1 - direct dependency for canonical serialization
(5/6) Wrote github.com/stretchr/testify@v1.3.0 - direct dependency
(6/6) Wrote github.com/oklog/ulid@v2.0.2 - direct dependency
```

To better assess this PR, I'll also create a branch for Duffle to see how it is affected.

Thoughts?


action/action_test.go
```diff
@@ -0,0 +1,238 @@
+package action
```
glyn: Now that we are effectively promoting these packages for reuse by others, I'd prefer black-box testing from the `action_test` package. Benefits include:
* the testcases are more likely to document the expected use of the package
* any test-specific externals of the package will stick out like a sore thumb, thus tending to reduce such externals.

claim/claim.go
```diff
@@ -0,0 +1,95 @@
...
+// Status constants define the CNAB status fields on a Result.
+const (
+	StatusSuccess  = "success"
+	StatusFailure  = "failure"
+	StatusUnderway = "underway"
+	StatusUnknown  = "unknown"
+)
+
+// Action constants define the CNAB action to be taken
+const (
```
glyn: Would it be worth introducing an `Action` type (based on `string`) to avoid actions being accidentally intermixed with raw strings? Similarly for `Status` above.

claim/claimstore.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/deislabs/cnab-go/utils/crud"
+)
+
+// ErrClaimNotFound represents a claim not found in claim storage
+var ErrClaimNotFound = errors.New(`Claim does not exist`)
```
glyn: nit: Why `` instead of "" to delimit the string?

silvin-lubecki:
After some thoughts, I wonder if this `crud` package should really be in the client-go for now 🤔 
I don't know what consumers can do with just this package... as is will need the claim store implementation to be able to persist claims like `duffle`. It may be practical as a pure dependency cleaning point of view, but from this repository point of view I don't see how it adds value.

radu-matei:
Yeah, totally agree that the name is not ideal - that being said, I didn’t want the initial PR to change anything else other than the import paths for Duffle (including @glyn’s comments).

Then, the only thing needed from the utils package is actually the Store interface.

So here’s my proposal - if the current state of this PR is acceptable, we work on updating Duffle  then do some more cleanup in a later PR - is this reasonable?
If not, I can surely update this PR with the changes mentioned. 

Thanks!

silvin-lubecki:
Sure let's merge this one and check how it feels from duffle and other consumers (I'll try to bump the dependency today on `docker app`). And this can totally be cleaned in a follow-up 👍

radu-matei:
Also opened https://github.com/deislabs/duffle/pull/742 in Duffle.

PR #15, opened 2019-05-10
added mongodb as a claim storage backend.

technosophos:
This adds a MongoDB backend for claims storage.

This is mainly intended to be an example of how to write claim storage backends.


technosophos:
Rebased.

PR #19, opened 2019-05-17
Convert action_tests into a black box test

### LINKED ISSUE #10: Black-box testing for the actions package
See [original comment from @glyn](https://github.com/deislabs/cnab-go/pull/8#discussion_r282618613):

> Now that we are effectively promoting these packages for reuse by others, I'd prefer black-box testing from the action_test package. Benefits include:
> 
> - the testcases are more likely to document the expected use of the package
> - any test-specific externals of the package will stick out like a sore thumb, thus tending to reduce such externals.

joaopapereira:
Play around with the idea of keeping the action tests by creating the tests in `action_test.go` and than call these tests for each command.

What y'all think about this idea?
These shared methods in action look very meaty so they should have some tests, but they will be all the same between action. This way we can reuse the tests between action

fixes #10

### DIFF
--- a/action/action_test.go
+++ b/action/action_test.go
@@ -1,18 +1,17 @@
-package action
+package action_test
 
 import (
-	"encoding/json"
 	"errors"
-	"os"
-	"strings"
+	"io/ioutil"
 	"testing"
 	"time"
 
+	"github.com/deislabs/cnab-go/action"
+	"github.com/deislabs/cnab-go/bundle"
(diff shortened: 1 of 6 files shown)


carolynvs:
@joaopapereira I apologize that this PR didn't get a review. Unfortunately at this point, the code has drifted enough that I don't think the PR is mergeable without a lot more work, so I'm closing this PR.

PR #21, opened 2019-05-31
Add custom extensions to bundle struct

carolynvs:
This adds the custom extensions section to Bundle from the CNAB Spec as part of the push to close gaps for 1.0. See https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#custom-extensions from the CNAB Spec.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -23,6 +23,10 @@ type Bundle struct {
 	Actions          map[string]Action              `json:"actions,omitempty" mapstructure:"actions"`
 	Parameters       map[string]ParameterDefinition `json:"parameters" mapstructure:"parameters"`
 	Credentials      map[string]Location            `json:"credentials" mapstructure:"credentials"`
+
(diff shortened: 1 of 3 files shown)


bundle/bundle.go
```diff
@@ -23,6 +23,10 @@ type Bundle struct {
 	Actions          map[string]Action              `json:"actions,omitempty" mapstructure:"actions"`
 	Parameters       map[string]ParameterDefinition `json:"parameters" mapstructure:"parameters"`
 	Credentials      map[string]Location            `json:"credentials" mapstructure:"credentials"`
+
+	// Custom extension metadata is a named collection of auxiliary data whose
+	// meaning is defined outside of the CNAB specification.
+	Custom map[string]interface{} `json:"custom" mapstructure:"custom"`
```
silvin-lubecki: Sorry for this late comment, but shouldn't we add a `omitempty` tag to the `custom` map, so it won't be serialized if empty ?

PR #23, opened 2019-06-03
Add missing omitempty to Bundle.Custom

carolynvs:
This is a follow-up to https://github.com/deislabs/cnab-go/pull/21/files#r289733051 to ensure that custom isn't serialized when empty.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -26,7 +26,7 @@ type Bundle struct {
 
 	// Custom extension metadata is a named collection of auxiliary data whose
 	// meaning is defined outside of the CNAB specification.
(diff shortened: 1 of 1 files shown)


carolynvs:
Yeah I was thinking the same thing. I'll submit one in a follow-up because I think that the PR may get nitted a bit (e.g. should it be canonical json? what other scenarios and variations should it test, etc), and I'd like to get this one in. 😀

radu-matei:
Regarding whether the tests should marshal bundles in canonical JSON form - I think they should, and FWIW, the methods for writing bundles ([`WriteFile` and `WriteTo`](https://github.com/deislabs/cnab-go/blob/master/bundle/bundle.go#L45-L63)) already do so in canonical form.

PR #24, opened 2019-06-03
Test bundle marshaling

carolynvs:
Ensure that the properties set on the bundle struct, such as omitempty, were set properly so that marshaling looks right.


bundle/testdata/minimal.json
```diff
@@ -0,0 +1,20 @@
...
+    }
+  ],
+  "images": {
+    "mybund": {
+      "imageType": "docker",
+      "image": "deislabs/mybund:abc123",
+      "description": "my bun daemon"
+    }
+  },
+  "parameters": null,
```
silvin-lubecki: Shouldn't these fields be removed 🤔 ?
I mean, should we add `omitempty` there too?

bundle/testdata/empty.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "",
+  "version": "",
+  "description": "",
+  "invocationImages": null,
```
silvin-lubecki: Shouldn't these fields be removed 🤔 ?
I mean, should we add omitempty there too?

bundle/bundle_test.go
```diff
@@ -246,3 +247,41 @@ func TestReadCustomExtensions(t *testing.T) {
 	assert.Equal(t, true, backupExt["enabled"])
 	assert.Equal(t, "daily", backupExt["frequency"])
 }
+
+func TestBundle_RoundTripEmptyJson(t *testing.T) {
+	wantData, err := ioutil.ReadFile("testdata/empty.json")
```
silvin-lubecki: You can merge those two tests and use the [TableDrivenTesting](https://github.com/golang/go/wiki/TableDrivenTests) pattern here, with something like this:

```go
func TestBundle_RoundTrip(t *testing.T){
    testCases := []struct{
        name string
        json string
    }{
        { name: "EmptyJson", json: "testdata/empty.json"},
        { name: "MinimalJson", json: "testdata/minimal.json"},
    }
    for _, tc := range testCases {
        t.Run(tc.name, func(t *testing.T) {
            wantData, err := ioutil.ReadFile(tc.json)
            ...
        }
    }
}
```

radu-matei:
As mentioned in #23, I think we should use the [`WriteFile` and `WriteTo`](https://github.com/deislabs/cnab-go/blob/master/bundle/bundle.go#L45-L63) methods for marshaling bundles, since they already to so in canonical form - thoughts?

bundle/testdata/empty.json
carolynvs: invocationImages is a required property, so I don't think omitEmpty makes sense for this field.

bundle/testdata/minimal.json
carolynvs: Agreed, this one should get an omitEmpty in a follow-up PR.

carolynvs:
Based on this, I think that the following fields should be flagged with omit empty in a follow-up PR because they are actually optional:

* parameters
* credentials
* outputs

carolynvs:
Ok, sorry it took me so long to turn around on your suggestions! I think that I've addressed them and this is ready for review again.

carolynvs:
Thanks to @glyn, this has been rebased and updated with a test for `schemaVersion` to make sure that it marshals and that it validates as a semver value.

bundle/bundle_test.go
```diff
@@ -190,17 +191,74 @@ func TestValidateVersionTag(t *testing.T) {
...
 	}
 
 	err := b.Validate()
 	is.EqualError(err, "'latest' is not a valid bundle version")
 }
 
+func TestValidateSchemaVersion(t *testing.T) {
+	is := assert.New(t)
+
+	img := InvocationImage{BaseImage{}}
```
silvin-lubecki: Totally nit picking, but I think `var img InvocationImage` is fine 😸

bundle/bundle_test.go
```diff
@@ -190,17 +191,74 @@ func TestValidateVersionTag(t *testing.T) {
...
+	b := Bundle{
+		Version:          "0.1.0",
+		SchemaVersion:    ".1",
+		InvocationImages: []InvocationImage{img},
+	}
+
+	err := b.Validate()
+	is.EqualError(err, "invalid bundle schema version \".1\": Invalid Semantic Version")
```
silvin-lubecki: Nit picking here too:
```go
`invalid bundle schema version ".1": Invalid Semantic Version`
```

carolynvs:
Okay, rebased!

PR #28, opened 2019-06-08
Proposal: refactor Operation API

jlegrone:
This PR introduces two new types, `EnvVar` and `File`. IMO this change helps make how to correctly use these objects more clear to end users of the public API.

This is similar to how [Container.Env](https://github.com/kubernetes/kubernetes/blob/c7972d9a5e7f555371fec7a3237f08ebc1ae9d7c/pkg/apis/core/types.go#L2030) is defined in kubernetes.

I've also removed `Parameters`, as I _think_ those should not be required here (they are already represented as environment variables and/or files).

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -23,8 +23,6 @@ type Operation struct {
 	Revision string `json:"revision"`
 	// Action is the action to be performed
 	Action string `json:"action"`
-	// Parameters are the parameters to be injected into the container
-	Parameters map[string]interface{} `json:"parameters"`
 	// Image is the invocation image
 	Image string `json:"image"`
 	// ImageType is the type of image.
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -38,17 +38,17 @@ type Operation struct {
 // EnvVar represents an environment variable set in the invocation image.
 type EnvVar struct {
 	// Name of the environment variable.
-	Name string
+	Name string `json:"name"`
 	// Value of the environment variable.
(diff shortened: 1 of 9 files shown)


driver/driver.go
```diff
@@ -23,20 +23,34 @@ type Operation struct {
...
-	Files map[string]string `json:"files"`
+	Files []File `json:"files"`
 	// Output stream for log messages from the driver
 	Out io.Writer
 }
 
+// EnvVar represents an environment variable set in the invocation image.
+type EnvVar struct {
+	// Name of the environment variable.
+	Name string
```
silvin-lubecki: nit: I think it misses the `json` annotation

driver/driver.go
```diff
@@ -23,20 +23,34 @@ type Operation struct {
...
+	// Value of the environment variable.
+	Value string
+}
+
+// A File represents the path and file content injected in the invocation image.
+type File struct {
+	// Path of the file, eg. /cnab/app/nginx.conf
+	Path string
+	// File content
+	Bytes []byte
```
silvin-lubecki: nit: What about `Content` instead?

jlegrone: I debated this and figured Bytes was a little more explicit, but I do prefer Content. It's in the type anyway. Switched.

jlegrone:
At this point there doesn't seem to be any interest and this would be a pretty sizeable breaking change. Closing.

PR #29, opened 2019-06-13
Updating credentials to use a struct with more than location

### LINKED ISSUE #27: Spec Change: Credentials now have a required attribute
The CNAB spec was updated to introduce a [required](https://github.com/deislabs/cnab-spec/pull/189) attribute on credentials. More generally, we don't properly model the credential as it also has a description.  We should update cnab-go to reflect that, and introduce a Credential struct that models this correctly, instead of just using the Location type.

jeremyrickard:
Per the spec, credentials have more attributes than just location, so this PR changes the definition around. Originally, this was just adding the `required` field but we also didn't have description. 

Closes #27

### DIFF
--- a/bundle/credentials.go
+++ b/bundle/credentials.go
@@ -1,0 +1,8 @@
+package bundle
+
+// Credential represents the definition of a CNAB credential
+type Credential struct {
+	Location    `json:",squash" mapstructure:",squash"`
+	Description string `json:"description,omitempty" mapstructure:"description"`
+	Required    bool   `json:"required,omitempty" mapstructure:"required"`
+}
--- a/bundle/credentials.go
+++ b/bundle/credentials.go
@@ -2,7 +2,7 @@ package bundle
 
 // Credential represents the definition of a CNAB credential
 type Credential struct {
-	Location    `json:",squash" mapstructure:",squash"`
+	Location    `mapstructure:",squash"`
 	Description string `json:"description,omitempty" mapstructure:"description"`
 	Required    bool   `json:"required,omitempty" mapstructure:"required"`
 }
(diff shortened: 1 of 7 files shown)


bundle/credentials.go
```diff
@@ -0,0 +1,8 @@
+package bundle
+
+// Credential represents the definition of a CNAB credential
+type Credential struct {
+	Location    `json:",squash" mapstructure:",squash"`
```
silvin-lubecki: nit: I guess this file needs a `go fmt` 😸

silvin-lubecki: Is the `squash` annotation available for `json` ? 🤔

jeremyrickard: Oops sorry, that's not needed :(

bundle/credentials.go
```diff
@@ -0,0 +1,8 @@
+package bundle
+
+// Credential represents the definition of a CNAB credential
+type Credential struct {
+	Location    `mapstructure:",squash"`
```
carolynvs: lol this sent me off on a rabbit hole of "huh, inlining doesn't use 'squash'..." then I found this for the mapstructure lib used by viper https://github.com/mitchellh/mapstructure/issues/42. It got my vote! 👍

bundle/credentials_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCompleteCredDefinition(t *testing.T) {
+	payload := `{
+		"credentials": {
+			"something": { 
+				"description" : "wicked this way comes",
```
carolynvs: 😎

PR #33, opened 2019-06-18
Delete originalImage

glyn:
Ref https://github.com/deislabs/cnab-spec/issues/195
Ref https://github.com/deislabs/duffle/pull/778


glyn:
Probably best to remove the uses of `originalImage` in `duffle` (e.g. by merging https://github.com/deislabs/duffle/pull/778) before merging this, in case anyone needs to make an urgent change here.

Issue #17, opened 2019-05-16
Update Parameter Definition To Match Spec


jeremyrickard:
Couple of outstanding issues with Parameter definitions:

* The top level parameters element of the bundle.json requires a required array to indicate what parameters are required for the bundle. 
* Simple types should not have a required attribute
* Our types are not correct (bool => boolean)
* JSON schema for complex objects

jeremyrickard:
This is closed by #34 and #30

Issue #14, opened 2019-05-10
support outputs in bundle pkg


michelleN:
as per the `bundle.json` [description](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md) and the [outputs PR](https://github.com/deislabs/cnab-spec/pull/133) on the cnab spec

radu-matei:
The parameters part of the specification also needs to be updated for the JSON schema, and the main question here is how to approach adding JSON schema support in all places where it is required, and also be consistent and _not reimplement_ the schema itself and the validation.

@carolynvs might have some thoughts about this, since she implemented parts of this for Porter.

jeremyrickard:
@radu-matei we added some json schema validation w/ the mixins but not at the bundle level yet (each of the mixins has it's own pre-known schema and we're using a library to validate that).

I think we can probably refactor part of that out and update cnab-go to have some common functions for dealing with bundle parameter definitions. I opened a separate issue to update to the rest of the parameter definition stuff as well (see: #17).

PR #34, opened 2019-06-19
Part 2: Introducing JSON Schema for Param and Outputs

jeremyrickard:
This is Part 2 of the change to introduce json schema support. This introduces a dependency on
https://github.com/qri-io/jsonschema and introduces a facade layer that hides that implementation detail
from the users of cnab-go. A struct was added, called Schema, to a new "definitions" package. This exposes an exported type that covers all the JSON schema bits. This was desirable to shield end users of cnab-go from the implementation details of a specific library and for some libraries, necessary because the generally used jsonschema libraries for Go either don't export the types directly (https://github.com/xeipuuv/gojsonschema), export but in a harder to directly use way (https://github.com/qri-io/jsonschema). A third library did check most of the boxes (https://github.com/santhosh-tekuri/jsonschema), but I felt like not exposing the choice directly to the end consumer makes it easier to switch if needed.

The wrapping package currently uses (https://github.com/qri-io/jsonschema and marshalls into that schema type when a validation is requested. This also removes old ParameterDefinition validation logic, as it is replaced by the use of json schema validation.

Part three will wire up all of this for outputs. They are currently present in structure, but there is no validation done in cnab-go for outputs (paths etc)

Part two of #17 and #14


jeremyrickard:
One thing to note here, our `definitions` section doens't match 100% with json schema:

https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#definitions


Notably: 

```
additionalItems: Parameter validation requiring that any additional items included in a user-provided array must conform to the specified schema. MUST be a JSON schema. (OPTIONAL)
additionalProperties: Parameter validation requiring that any additional properties in the user-provided object conform to the specified schema. MUST be a JSON schema. (OPTIONAL)
```

These differ from schema. `additionalProperties`, for example can be a boolean or schema to define additional types. I went with schema here, but can restrict it down.

bundle/definition/schema.go
```diff
@@ -0,0 +1,120 @@
...
+// UnmarshalJSON provides an implementation of a JSON unmarshaler that uses the
+// github.com/qri-io/jsonschema to load and validate a given schema. If it is valid,
+// then the json is unmarshaled.
+func (s *Schema) UnmarshalJSON(data []byte) error {
+	js := new(jsonschema.RootSchema)
+	err := js.UnmarshalJSON(data)
+	if err != nil {
+		return err
+	}
+	type McGuffin Schema
```
carolynvs: Looks like some test code got left behind

bundle/definition/validation.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/qri-io/jsonschema"
+)
+
+// Validate applies JSON Schema validation to the data passed as a paraemeter.
```
carolynvs: s/paraemeter/parameter

bundle/definition/validation.go
```diff
@@ -0,0 +1,69 @@
...
+	if len(valErrs) > 0 {
+		var anError error
+		for _, err := range valErrs {
+			if anError == nil {
+				anError = errors.New(fmt.Sprintf("unable to validate %s, error: %s", err.PropertyPath, err.Message))
+			} else {
+				anError = errors.Wrap(anError, fmt.Sprintf("unable to validate %s, error: %s", err.PropertyPath, err.Message))
+			}
+		}
+		return errors.Wrap(anError, "invalid parameter or output")
```
carolynvs: This is going to be incredibly difficult for a person to figure out what's wrong with their bundle, and maybe impossible for a tool because it's not structured. Long term, do you think we could return a structured message that tells them exactly what is wrong with property so that they can examine it and unpack each property and validation error?

bundle/outputs.go
```diff
@@ -1,20 +1,13 @@
...
-	MaxLength     *int               `json:"maxLength,omitempty" mapstructure:"maxLength"`
-	Metadata      *ParameterMetadata `json:"metadata,omitempty" mapstructure:"metadata"`
-	Path          string             `json:"path,omitemtpty" mapstructure:"path,omitempty"`
-	ApplyTo       []string           `json:"apply-to,omitempty" mapstructure:"apply-to,omitempty"`
-	Sensitive     bool               `json:"sensitive,omitempty" mapstructure:"sensitive,omitempty"`
+	Definition  string   `json:"definition" mapstructure:"definition"`
+	ApplyTo     []string `json:"applyTo,omitempty" mapstructure:"applyTo,omitempty"`
+	Description string   `json:"description,omitempty" mapstructure:"description"`
+	Path        string   `json:"path,omitemtpty" mapstructure:"path,omitempty"`
+	Sensitive   bool     `json:"sensitive,omitempty" mapstructure:"sensitive,omitempty"`
```
carolynvs: Did we change our minds about `writeOnly` representing sensitive? Looks like only outputs have this but on ParameterDefinition.

bundle/definition/validation.go
jeremyrickard: Yeah, can change to that.

bundle/outputs.go
jeremyrickard: oops, good catch! that just got changed in the spec.

bundle/definition/schema.go
jeremyrickard: I've added docs to clarify this.

bundle/definition/validation.go
jeremyrickard: Refactored this part. Now returns an slice of ValidationErrors{ Path string, Error string}

bundle/bundle.go
```diff
@@ -8,23 +8,26 @@ import (
...
+	Description      string                 `json:"description" mapstructure:"description"`
+	Keywords         []string               `json:"keywords,omitempty" mapstructure:"keywords"`
+	Maintainers      []Maintainer           `json:"maintainers,omitempty" mapstructure:"maintainers"`
+	InvocationImages []InvocationImage      `json:"invocationImages" mapstructure:"invocationImages"`
+	Images           map[string]Image       `json:"images" mapstructure:"images"`
+	Actions          map[string]Action      `json:"actions,omitempty" mapstructure:"actions"`
+	Parameters       ParametersDefinition   `json:"parameters" mapstructure:"parameters"`
+	Credentials      map[string]Credential  `json:"credentials" mapstructure:"credentials"`
+	Outputs          OutputsDefinition      `json:"outputs" mapstructure:"outputs"`
+	Definitions      definition.Definitions `json:"definitions" mapstructure:"definitions"`
```
silvin-lubecki: Is this section mandatory or optional? Should it be tagged as `omitempty` ?

bundle/bundle.go
```diff
@@ -139,17 +142,28 @@ func ValuesOrDefaults(vals map[string]interface{}, b *Bundle) (map[string]interf
...
+		s, ok := b.Definitions[def.Definition]
+		if !ok {
+			return res, fmt.Errorf("unable to find definition for %s", name)
+		}
 		if val, ok := vals[name]; ok {
-			if err := def.ValidateParameterValue(val); err != nil {
-				return res, fmt.Errorf("can't use %v as value of %s: %s", val, name, err)
+			valErrs, err := s.Validate(val)
+			if err != nil {
+				return res, pkgErrors.Wrap(err, fmt.Sprintf("encountered an error validating parameter %s", name))
```
silvin-lubecki: nit: If I remember correctly, there's a `pkgErrors .Wrapf` function.

bundle/definition/schema.go
```diff
@@ -0,0 +1,126 @@
...
+	Required      []string           `json:"required,omitempty" mapstructure:"required,omitempty"`
+	Then          *Schema            `json:"then,omitempty" mapstructure:"then,omitempty"`
+	Title         string             `json:"title,omitempty" mapstructure:"title,omitempty"`
+	Type          interface{}        `json:"type,omitempty" mapstructure:"type,omitempty"`
+	UniqueItems   *bool              `json:"uniqueItems,omitempty" mapstructure:"uniqueItems,omitempty"`
+	WriteOnly     *bool              `json:"writeOnly,omitempty" mapstructure:"writeOnly,omitempty"`
+}
+
+// SingleType returns true if a schema has a single type declaration
+func (s *Schema) SingleType() bool {
```
silvin-lubecki: Is this function needs to be exported? It feels like it will only be used by `GetType` and `GetTypes` and not by end users who would mostly use `GetType` (but maybe I'm wrong).

bundle/definition/schema.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	data, ok := s.Type.([]interface{})
+	if !ok {
+		return nil, false, errors.New("unknown type values")
+	}
+	typeStrings := []string{}
+	for _, val := range data {
+		typeString, ok := val.(string)
+		if !ok {
+			return nil, false, errors.New("unknown type value")
```
silvin-lubecki: nit: add the `val` type to the error message?

bundle/definition/schema.go
```diff
@@ -0,0 +1,126 @@
...
+// UnmarshalJSON provides an implementation of a JSON unmarshaler that uses the
+// github.com/qri-io/jsonschema to load and validate a given schema. If it is valid,
+// then the json is unmarshaled.
+func (s *Schema) UnmarshalJSON(data []byte) error {
+
+	// Before we unmarshal into the cnab-go bundle/definition/Schema type, unmarshal into
+	// the library struct so we can handle any validation errors in the schema. If there
+	// are any errors, return those.
+	js := new(jsonschema.RootSchema)
+	err := js.UnmarshalJSON(data)
```
silvin-lubecki: nit: `if err := js.UnmarshalJSON(data); err != nil {`

bundle/definition/schema.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	// The schema is valid at this point, so now use an indirect wrapper type to actually
+	// unmarshal into our type.
+	type wrapperType Schema
+	wrapper := struct {
+		*wrapperType
+	}{
+		wrapperType: (*wrapperType)(s),
+	}
+	err = json.Unmarshal(data, &wrapper)
```
silvin-lubecki: nit: `return json.Unmarshal(data, &wrapper)`

bundle/definition/schema.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return err
+	}
+	// The schema is valid at this point, so now use an indirect wrapper type to actually
+	// unmarshal into our type.
+	type wrapperType Schema
+	wrapper := struct {
+		*wrapperType
+	}{
+		wrapperType: (*wrapperType)(s),
```
silvin-lubecki: Is this trick used to avoid the `json.Unmarshal` function to call this very method `func (s *Schema) UnmarshalJSON`, and get stuck into an infinite recursive call?

bundle/parameters_test.go
```diff
@@ -3,8 +3,7 @@ package bundle
 import (
 	"fmt"
 	"testing"
-
-	"github.com/stretchr/testify/assert"
+	//"github.com/stretchr/testify/assert"
```
silvin-lubecki: nit: commented code

bundle/definition/schema.go
jeremyrickard: Indeed it is.

bundle/bundle.go
jeremyrickard: You are correct! I'm not sure why I forgot that, thanks for the reminder!

bundle/bundle.go
jeremyrickard: I'm not sure. Parameters and Outputs are also not included, but I think all three probably end up being non-required. I'll tag here and we can revisit the other two in a specific PR/issue combo?

bundle/parameters_test.go
jeremyrickard: Thank you!

bundle/definition/schema.go
jeremyrickard: 👍

bundle/definition/schema.go
jeremyrickard: 👍

bundle/definition/schema.go
```diff
@@ -0,0 +1,121 @@
...
+	return ok
+}
+
+// GetType will return the singular type for a given schema and a success boolean. If the
+// schema does not have a single type, it will return the false boolean and an error.
+func (s *Schema) GetType() (string, bool, error) {
+	if !s.singleType() {
+		return "", false, errors.New("this schema has multiple types")
+	}
+	typeString, ok := s.Type.(string)
```
silvin-lubecki: I wonder if the previous call to `singleType` could not be directly handled here, as the cast here is exactly the one made by `singleType` (feels like we are doing the same thing twice). Could it be simplified with something like this?
```go
func (s *Schema) GetType() (string, bool, error) {
    typeString, ok := s.Type.(string)
    if !ok {
        return "", false, errors.New("this schema has multiple types")
    }
    return typeString, ok, nil
}
```

jeremyrickard: You're correct, when I un-exported the other method I should have just done this.

bundle/definition/schema.go
```diff
@@ -0,0 +1,109 @@
...
+	UniqueItems   *bool              `json:"uniqueItems,omitempty" mapstructure:"uniqueItems,omitempty"`
+	WriteOnly     *bool              `json:"writeOnly,omitempty" mapstructure:"writeOnly,omitempty"`
+}
+
+// GetType will return the singular type for a given schema and a success boolean. If the
+// schema does not have a single type, it will return the false boolean and an error.
+func (s *Schema) GetType() (string, bool, error) {
+	typeString, ok := s.Type.(string)
+	if !ok {
+		return "", false, errors.New("couldn't get type for schema")
```
silvin-lubecki: What about something like: `fmt.Errorf("this schema has multiple types: %t", typeString)` ?

PR #39, opened 2019-06-20
Update bundle.go to support empty images, parameters, creds and outputs

### LINKED ISSUE #38: Add `omitempty` to OPTIONAL fields
There are some fields in the bundle that are optional per the CNAB spec:

* Images
* Credentials
* Parameters
* Outputs

We should update the bundle.go code appropriately to ensure that these can be left out of the resulting JSON without things like:

```
{"credentials":null,"parameters":null}
```

jeremyrickard:
This adds an "omitempty" tag to images, credentials, parameters and outputs. ParametersDefinition and OutputsDefinition also become pointers, so that they aren't needlessly output as empty structs. 

Closes #38


bundle/bundle.go
```diff
@@ -23,11 +23,11 @@ type Bundle struct {
 	Keywords         []string               `json:"keywords,omitempty" mapstructure:"keywords"`
 	Maintainers      []Maintainer           `json:"maintainers,omitempty" mapstructure:"maintainers"`
 	InvocationImages []InvocationImage      `json:"invocationImages" mapstructure:"invocationImages"`
-	Images           map[string]Image       `json:"images" mapstructure:"images"`
-	Actions          map[string]Action      `json:"actions,omitempty" mapstructure:"actions"`
-	Parameters       ParametersDefinition   `json:"parameters" mapstructure:"parameters"`
-	Credentials      map[string]Credential  `json:"credentials" mapstructure:"credentials"`
-	Outputs          OutputsDefinition      `json:"outputs" mapstructure:"outputs"`
+	Images           map[string]Image       `json:"images,omitempty" mapstructure:"images,omitempty"`
```
carolynvs: mapstructure doesn't support omitempty, so the extra tag should only go on json.

jeremyrickard: ahhh ok!

jeremyrickard: removed

PR #40, opened 2019-06-20
Updated param and output fields to also have an omitempty

jeremyrickard:
Adds "omitempty" to both ParametersDefinition.Fields and OutputsDefinition.Fields

### DIFF
--- a/bundle/outputs.go
+++ b/bundle/outputs.go
@@ -1,7 +1,7 @@
 package bundle
 
 type OutputsDefinition struct {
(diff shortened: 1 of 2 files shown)


jeremyrickard:
I agree with @carolynvs. I think that *when* the struct is present, then `fields` is required. The spec doesn't say that, but I think it's implicit.

silvin-lubecki:
Right, agree with @carolynvs too. @jeremyrickard do you want to close this PR then?

jeremyrickard:
Closing per agreement above.

PR #41, opened 2019-06-20
Reintroduce ConvertValue(interface{})

jeremyrickard:
This PR re-introduces the ConvertValue(interface{}) method, but relocates it to
the definitions package.

The method was removed during the refactor since the types can be more than the simple types before, but was added back for convenience.  It only works on number, integer, boolean or string types. Others return an error.

### DIFF
--- a/bundle/definition/schema.go
+++ b/bundle/definition/schema.go
@@ -2,6 +2,8 @@ package definition
 
 import (
 	"encoding/json"
+	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/qri-io/jsonschema"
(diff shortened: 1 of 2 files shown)


bundle/definition/schema.go
```diff
@@ -107,3 +109,34 @@ func (s *Schema) UnmarshalJSON(data []byte) error {
...
+	case "number":
+		num, err := strconv.ParseFloat(val, 64)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to convert %s to number", val)
+		}
+		return num, nil
+	case "integer":
+		return strconv.Atoi(val)
+	case "boolean":
+		if strings.ToLower(val) == "true" {
```
silvin-lubecki: nit: what about ?
```go
switch strings.ToLower(val){
case "true":
    return true, nil
case "false":
    return false, nil
default:
    return false, errors.Errorf("%q is not a valid boolean", val)
}
```

jeremyrickard: Hey @silvin-lubecki I can do a follow-up for this. This was basically just re-introducing the existing code. I like your suggestion though, so I will open a followon.

silvin-lubecki:
Arg, just made a comment the moment you merged @carolynvs 😅

carolynvs:
Sorry about that! I had 2 reviews and it seemed uncontroversial to add back accidentally deleted code... 😊  Looks like Jeremy is going to address your comment in a follow-up PR.

PR #43, opened 2019-06-20
Add Image relocation Map to translate invocation image

joaopapereira:
This PR start the solution for the issue: deislabs/duffle#785

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -36,6 +36,11 @@ func selectInvocationImage(d driver.Driver, c *claim.Claim) (bundle.InvocationIm
 
 	for _, ii := range c.Bundle.InvocationImages {
 		if d.Handles(ii.ImageType) {
+			if c.RelocationMap != nil {
+				if img, ok := c.RelocationMap[ii.Image]; ok {
+					ii.Image = img
+				}
+			}
 			return ii, nil
 		}
 	}
--- a/claim/claim.go
+++ b/claim/claim.go
@@ -35,14 +35,15 @@ const (
 // provide the necessary data to upgrade, uninstall, and downgrade
 // a CNAB package.
 type Claim struct {
-	Name       string                 `json:"name"`
-	Revision   string                 `json:"revision"`
-	Created    time.Time              `json:"created"`
-	Modified   time.Time              `json:"modified"`
-	Bundle     *bundle.Bundle         `json:"bundle"`
-	Result     Result                 `json:"result"`
-	Parameters map[string]interface{} `json:"parameters"`
-	Files      map[string]string      `json:"files"`
+	Name          string                    `json:"name"`
+	Revision      string                    `json:"revision"`
+	Created       time.Time                 `json:"created"`
+	Modified      time.Time                 `json:"modified"`
+	Bundle        *bundle.Bundle            `json:"bundle"`
+	Result        Result                    `json:"result"`
+	Parameters    map[string]interface{}    `json:"parameters"`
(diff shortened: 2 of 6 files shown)


action/action.go
```diff
@@ -26,16 +26,21 @@ const stateful = false
 // - status
 type Action interface {
 	// Run an action, and record the status in the given claim
-	Run(*claim.Claim, credentials.Set, io.Writer) error
+	Run(*claim.Claim, credentials.Set, bundle.ImageRelocationMap, io.Writer) error
```
silvin-lubecki: I wonder if maybe this relocation map should be stored in the Claim too 🤔

joaopapereira: I thought about that as well and it was going to be my first approach bu I realized that the relocation map is a parameter similar to the creds file.

The current process for relocation would be:
`duffle relocate` that will move the images a new registry, this is also the step that generations the relocation map
`duffle import` that will create the claim(we need this at this point because there is not support for installing thick bundle)
`duffle install` when we pass in the registry map

The advantage I see here is that if it behaves like the creds we can maybe relocate and then upgrade an installation. If we save it in the claims and relocate the bundle the claims will be out of date and the new mapping file will have to be passed.
The downside here is that we will have to pass on all the action the mapping and we need the operator would have to store that file somewhere.

If we move it to claims we would only need to pass the mapping on import/install. The drawback is that on upgrade we would have to supply the mapping and update the claim as well.

Nevertheless I am open to the change, what do y'll think?

silvin-lubecki: I think I prefer having it in the claim, as for the user this relocation file is somehow an implementation detail.

I guess it would be simpler for the user if the `import` command creates the claim with the relocation inside, so the user won't have to persist it manually or to remember which relocation file is linked with which CNAB bundle.

And about the upgrade, I think that a user upgrading a CNAB package with another one relocated somewhere else is (or should) not be a common use case (it could be seen as installing a brand new bundle).

joaopapereira: As per our conversation I moved the relocation map into the claims

claim/claim.go
```diff
@@ -35,21 +35,22 @@ const (
...
+	Files         map[string]string         `json:"files"`
+	RelocationMap bundle.ImageRelocationMap `json:"relocationMap"`
 }
 
 // ValidName is a regular expression that indicates whether a name is a valid claim name.
 var ValidName = regexp.MustCompile("^[a-zA-Z0-9_-]+$")
 
 // New creates a new Claim initialized for an installation operation.
-func New(name string) (*Claim, error) {
+func New(name string, relocationMap bundle.ImageRelocationMap) (*Claim, error) {
```
silvin-lubecki: I'm a bit on the fence for this, as adding the relocationMap here it is not homogeneous with other fields (like `parameters` or even `bundle`) which are not initialized in this `New` function.
What about removing it from here and let the user set the relocation map herself afterwards?

technosophos: I'm with @silvin-lubecki on this. Unless having a non-empty relocation map is _mandatory_, I would prefer to not require it during construction.

action/action.go
technosophos: That looks good. Thanks for doing that.

claim/claim.go
joaopapereira: I was unsure as well when I added it. From the side we only call `New` on install so I assume that we could piggy back on it. Nevertheless it is a better patter to do not have it there.

Just updated the PR and remove the map from the `New` method

PR #48, opened 2019-06-21
exclude Out from json marshalling in Operation

### LINKED ISSUE #47: Operation struct includes Out field when marshaled to son
The Out field is an io.writer, including it in json serialization means that CNAB-go cannot be used to implement a duffle command driver as unmarshall of {} fails.

simongdavies:
fixes #47


jeremyrickard:
@simongdavies agree with @silvin-lubecki. Could you add some tests around this?

PR #50, opened 2019-06-25
Fix a nil pointer deference in bundle.ValuesOrDefaults

jcsirot:
bundle.ValuesOrDefaults failed with a panic due to a nil pointer dereference when no parameters are defined in the bundle passed to the function.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -140,11 +140,16 @@ type Action struct {
 
 // ValuesOrDefaults returns parameter values or the default parameter values
 func ValuesOrDefaults(vals map[string]interface{}, b *Bundle) (map[string]interface{}, error) {
+	res := map[string]interface{}{}
+
+	if b.Parameters == nil {
+		return res, nil
+	}
+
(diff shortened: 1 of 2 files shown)


bundle/bundle.go
```diff
@@ -140,35 +140,38 @@ type Action struct {
...
-	for _, key := range b.Parameters.Required {
-		requiredMap[key] = struct{}{}
-	}
 	res := map[string]interface{}{}
 
-	for name, def := range b.Parameters.Fields {
-		s, ok := b.Definitions[def.Definition]
-		if !ok {
-			return res, fmt.Errorf("unable to find definition for %s", name)
+	if b.Parameters != nil {
```
silvin-lubecki: It's nit picking, but I prefer the early return notation:
```go
if b.Parameters == nil{
    return res, nil
}
...
```

jcsirot: I agree, here it is probably better for code (and PR) readability. Fixed.

PR #51, opened 2019-06-25
Update readme

joaopapereira:
- Add a little more information into the readme.
- Add goland ide files to gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,4 +4,58 @@ cnab-go is a library for building [CNAB](https://github.com/deislabs/cnab-spec) 
 
 cnab-go is currently being used by [Docker App](https://github.com/docker/app), [Duffle](https://github.com/deislabs/duffle), and [Porter](https://github.com/deislabs/porter). If you'd like to see your CNAB project listed here, please submit a PR.
 
+### Community
+
 cnab-go is [maintained](GOVERNANCE.md) by the CNAB community. We sometimes discuss cnab-go issues during the [bi-weekly CNAB community  meeting](https://hackmd.io/s/SyGcBcwQ4), but we encourage open communication via our [issue](https://github.com/deislabs/cnab-go/issues) queue and via [PRs](https://github.com/deislabs/cnab-go/pulls). If you are interested in contributing to cnab-go, please refer to our [contributing](CONTRIBUTING.md) guidelines.
+
+### Development
+
+#### Getting the code
+
+Cloning this repository and change directory to it:
+```bash
+$ go get -d github.com/deislabs/cnab-go/...
+$ cd $(go env GOPATH)/src/github.com/deislabs/cnab-go
+```
+
+#### Prerequisites
+
+You need:
+
+* make
+* Go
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,4 +4,58 @@ cnab-go is a library for building [CNAB](https://github.com/deislabs/cnab-spec)
...
 cnab-go is [maintained](GOVERNANCE.md) by the CNAB community. We sometimes discuss cnab-go issues during the [bi-weekly CNAB community  meeting](https://hackmd.io/s/SyGcBcwQ4), but we encourage open communication via our [issue](https://github.com/deislabs/cnab-go/issues) queue and via [PRs](https://github.com/deislabs/cnab-go/pulls). If you are interested in contributing to cnab-go, please refer to our [contributing](CONTRIBUTING.md) guidelines.
+
+### Development
+
+#### Getting the code
+
+Cloning this repository and change directory to it:
+```bash
+$ go get -d github.com/deislabs/cnab-go/...
+$ cd $(go env GOPATH)/src/github.com/deislabs/cnab-go
```
jlegrone: This will change as soon as #52 is merged. Perhaps we should get that in first?

radu-matei: Maybe not be so specific about where to clone it?

README.md
```diff
@@ -4,4 +4,58 @@ cnab-go is a library for building [CNAB](https://github.com/deislabs/cnab-spec)
...
+
+Retrieve all needed packages to start developing.
+This will download the binaries for the linter, dep and go imports in the end it will
+run `dep ensure` to download all the go package dependencies
+
+```bash
+$ make bootstrap
+```
+
+#### Building
```
radu-matei: Personal preference here - maybe have a single section for building, testing, and linting?

README.md
joaopapereira: I think I agree with @jlegrone that maybe this change here should wait for #52 to be merged before merging this one.
And when we that happens we can remove some of these lines 😄

jeremyrickard:
@jlegrone @joaopapereira since we're going to hold off on the go.mod support, do ya'll feel like we can merge this now w/ that in mind and we can come back and update this? I think overall the updates are good to have

joaopapereira:
@jeremyrickard I am ok merging it is as is, if @radu-matei is ok.

.gitignore
```diff
@@ -6,3 +6,4 @@ vendor/
 __pycache__/
 /duffle
 .vscode/settings.json
+.idea
```
glyn: There is a school of thought that developers should be responsible for ignoring their own IDE's settings in a global gitignore file so as not to pollute the gitignore files of multiple projects. However, this change seems worthwhile because goland (and vscode) are so prevalent.

silvin-lubecki: I agree with this school of thought 👍

joaopapereira:
I addressed the comments except the location where the code gets pulled to. For now that is the correct location, correct?

PR #53, opened 2019-06-26
Import drivers from duffle

### LINKED ISSUE #49: Migrate duffle driver implementations to cnab-go
The driver interface is already defined in the [driver](https://github.com/deislabs/cnab-go/blob/8507768ec67ce0d427f7b71f784eda67ffbe94f6/driver/driver.go) package. Many of the [driver implementations from duffle](https://github.com/deislabs/duffle/tree/e50cd9ec76e31da9126b18aa7a0e0eedd379cc1c/pkg/driver) are generally useful outside of duffle and should be made available without the need for declaring a dependency on duffle.

jlegrone:
Closes #49


jlegrone:
> I think we should put each driver in its own package

@silvin-lubecki I agree -- this will also help with potential for global variable/function naming collisions within the driver package. I'll tackle this later today.

jlegrone:
@silvin-lubecki I do need to go back and fix dependencies (will force push something in a few minutes). I've preserved the git history from duffle for the relevant directories so that git blame/log continues to attribute work to the correct authors.

radu-matei:
I'm also not sure we want all the history here.

jlegrone:
I would rather not be seen as the docker driver expert, so the main motivation in keeping history was to continue attributing lines to original authors, ie:

https://github.com/deislabs/cnab-go/blame/31f60ac6680d8f09bb5daf2fc64d7c4dcdc4dceb/driver/docker/docker.go

Sometimes this stuff provides really useful context for why something ended up a certain way, or who you should reach out to for review of a PR.

jlegrone:
Please don't merge yet; I'd first like to try modifying the duffle commit messages referencing PRs and issues to point back to the duffle repo rather than conflicting with this repo, which could be confusing.

silvin-lubecki:
^ @glyn any thought on that ?

glyn:
> ^ @glyn any thought on that ?

In general I'm against rewriting history in a git repo, but I take the point about correctly attributing credit/blame, so I agree with it in this case.

silvin-lubecki:
OK then, let's keep the history here 👍

PR #54, opened 2019-06-27
exclude Out from json marshalling in Operation

### LINKED ISSUE #47: Operation struct includes Out field when marshaled to son

simongdavies:
fixes #47.

@jeremyrickard , @silvin-lubecki this replace #48 , now has marshall/unmarshall tests

### DIFF
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -1,7 +1,10 @@
 package driver
 
 import (
+	"encoding/json"
 	"io/ioutil"
+	"os"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


driver/driver_test.go
```diff
@@ -30,3 +32,49 @@ func TestDebugDriver_Run(t *testing.T) {
...
+	var op Operation
+	is := assert.New(t)
+	bytes, err := ioutil.ReadFile("../testdata/operations/valid-operation.json")
+	is.NoError(err, "Error reading from testdata/operations/valid-operation.json")
+	is.NoError(json.Unmarshal(bytes, &op), "Error unmarshalling operation")
+	is.NotNil(op, "Expected Operation not to be nil")
+}
+
+func TestOperation_Marshall(t *testing.T) {
+
```
silvin-lubecki: nit: empty line

driver/driver_test.go
```diff
@@ -30,3 +32,49 @@ func TestDebugDriver_Run(t *testing.T) {
...
 	is.NoError(d.Run(op))
 }
+
+func TestOperation_Unmarshall(t *testing.T) {
+	var op Operation
+	is := assert.New(t)
+	bytes, err := ioutil.ReadFile("../testdata/operations/valid-operation.json")
+	is.NoError(err, "Error reading from testdata/operations/valid-operation.json")
+	is.NoError(json.Unmarshal(bytes, &op), "Error unmarshalling operation")
+	is.NotNil(op, "Expected Operation not to be nil")
```
silvin-lubecki: Could you test with a `DeepEqual` instead? So you are sure the unmarshalled struct is valid.

driver/driver_test.go
```diff
@@ -30,3 +32,49 @@ func TestDebugDriver_Run(t *testing.T) {
...
+		},
+		Out: os.Stdout,
+	}
+	is := assert.New(t)
+	bytes, err := json.Marshal(actualOp)
+	is.NoError(err, "Error Marshalling actual operation to json")
+	is.NotNil(bytes, "Expected marshalled json not to be nil")
+	actualJSON := string(bytes)
+	var expectedOp Operation
+	bytes, err = ioutil.ReadFile("../testdata/operations/valid-operation.json")
```
silvin-lubecki: Why don't you just check the equality between those bytes and the actualJSON bytes?

simongdavies: I did it this way so that I don't have to worry about the formatting of the json, if the way that operation is serialized changes then this will still work. I guess I could have minified the json and compared those bytes but this didn't require any additional libs.

carolynvs: Speaking as someone who just had to work with a test in cnab-go that asserted on the bytes instead of the json string representation, comparing strings instead of bytes is also a lot easier to deal with when the test fails or needs to be updated.

PR #55, opened 2019-07-09
Add dependencies to bundle struct

### LINKED ISSUE #35: Introduce Dependencies to Bundle
We should add support for [Section 500](https://github.com/deislabs/cnab-spec/blob/master/500-CNAB-dependencies.md).

carolynvs:
This adds the dependencies metadata from the [Dependencies Spec](https://github.com/deislabs/cnab-spec/blob/master/500-CNAB-dependencies.md) to the bundle struct.

Closes #35


silvin-lubecki:
@carolynvs I am wondering if this PR is maybe a little premature, as the specification is not yet merged and no runtime is able to manage dependencies yet.
What do you think about running an experiment on this subject (maybe on a draft branch on duffle or porter?), using the `Custom` field in the `Bundle` struct?
I guess that once a first version (like a Proof of Concept) of the runtime handling dependencies is ready, we'll have a clearer view of what should be put in the dependencies struct. I just fear that dependency management is a huge beast and we'll have to iterate several times on it, so why don't we iterate out of this repo and merge it when it's ready?

Does that make sense ?

carolynvs:
The spec was merged months ago actually, and at the time this was necessary in order to write to bundle.json so that I _could_ implement it in a runtime. 😀  

However since in the meeting yesterday we decided to move the dependencies section under custom,  this PR is no longer needed, so I am closing this. Just didn't get around to it in time. Sorry for the confusion!

PR #58, opened 2019-07-12
Updated the BaseImage definition to replace platform with Labels

jeremyrickard:
Fixes #58


silvin-lubecki:
Oh just wondering, is the type `ImagePlatform` still used? https://github.com/deislabs/cnab-go/blob/5ca85500e801be695748cf1f360f2254f425d19c/bundle/bundle.go#L89
Maybe should we remove it if it's not the case?

jeremyrickard:
> Oh just wondering, is the type `ImagePlatform` still used?
> 
> https://github.com/deislabs/cnab-go/blob/5ca85500e801be695748cf1f360f2254f425d19c/bundle/bundle.go#L89
> 
> 
> Maybe should we remove it if it's not the case?

YES, good catch @silvin-lubecki

PR #63, opened 2019-07-18
Adds command driver to cnab-go

simongdavies:
Moves command driver into its own package to match other drivers
Changes naming pattern for a driver from duffle-<driver-name> to cnab-<driver-name>


driver/command/command.go
```diff
@@ -0,0 +1,105 @@
...
+	// environment and adding in all the extra env vars.
+	pairs := os.Environ()
+	added := []string{}
+	for k, v := range op.Environment {
+		pairs = append(pairs, fmt.Sprintf("%s=%s", k, v))
+		added = append(added, k)
+	}
+	// DUFFLE_VARS is a list of variables we added to the env. This is to make
+	// it easier for shell script drivers to clone the env vars.
+	pairs = append(pairs, fmt.Sprintf("DUFFLE_VARS=%s", strings.Join(added, ",")))
```
vdice: Should we change `DUFFLE_VARS` to a different name?  Or, is this still useful to keep around?

simongdavies: Good catch, updated it.

PR #64, opened 2019-07-19
Handle immutable parameters in Bundle.ValuesOrDefaults

jcsirot:
This PR introduces a breaking change in the Bundle.ValuesOrDefaults function signature by taking a map of current parameter values in order to handle the recently introduced `immutable` field in the bundle parameters description.

The patch verifies when a parameter value is set:
1. the parameter has the immutable field set in the bundle
2. there is a previously set value for this parameter (i.e. a value has been set during the initial installation of the bundle) 
3. the new and existing values of this parameter are not identical

In that case an error is returned.

### DIFF
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -247,7 +247,7 @@ func TestValuesOrDefaults_Immutable(t *testing.T) {
 	}
 
 	_, err := ValuesOrDefaults(vals, currentVals, b)
-	is.Error(err)
+	is.EqualError(err, "parameter namespace is immutable and cannot be overridden with value new-ns")
 }
 
 func TestValidateVersionTag(t *testing.T) {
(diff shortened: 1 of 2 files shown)


bundle/bundle_test.go
```diff
@@ -205,11 +208,48 @@ func TestValuesOrDefaults_Required(t *testing.T) {
...
+				},
+				"enabled": {
+					Definition: "enabledType",
+				},
+			},
+		},
+	}
+
+	_, err := ValuesOrDefaults(vals, currentVals, b)
+	is.Error(err)
```
silvin-lubecki: What about testing the returned error is due to immutability? With something like this:
https://github.com/stretchr/testify/blob/ffdc059bfe9ce6a4e144ba849dbedead332c6053/assert/assertion_forward.go#L137 ?

carolynvs: Agreed, checking the error returned would be good

jcsirot: Indeed. I fixed it.

PR #66, opened 2019-07-19
feat(claim.go): add Outputs field to Claim struct

vdice:
- add Outputs field to Claim struct, per cnab-spec (updated in https://github.com/deislabs/cnab-spec/pull/217)

### DIFF
--- a/claim/claimstore_test.go
+++ b/claim/claimstore_test.go
@@ -118,3 +118,48 @@ func TestReadAll(t *testing.T) {
 	is.Equal("bar", claim2.Name)
 	is.Equal("baz", claim3.Name)
 }
+
+func TestCanUpdateOutputs(t *testing.T) {
+	is := assert.New(t)
+	claim, err := New("foo")
+	is.NoError(err)
+	is.Equal(map[string]interface{}{}, claim.Outputs)
+
+	tempDir, err := ioutil.TempDir("", "cnabgotest")
+	is.NoError(err, "Failed to create temp dir")
+	defer os.RemoveAll(tempDir)
+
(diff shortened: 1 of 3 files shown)


claim/claimstore_test.go
```diff
@@ -118,3 +118,49 @@ func TestReadAll(t *testing.T) {
...
+	is.NoError(err, "Failed to store claim")
+
+	c, err := store.Read("foo")
+	is.NoError(err, "Failed to read claim")
+
+	want := map[string]interface{}{
+		"foo-output": true,
+		"bar-output": "bar",
+	}
+	is.Equal(c.Outputs, want, "Wrong outputs on claim")
```
carolynvs: The signature for Equal is first the expected result, then the actual result. This got switched here which can lead to confusing error messages if the test were to fail.

claim/claimstore_test.go
```diff
@@ -118,3 +118,49 @@ func TestReadAll(t *testing.T) {
...
+	is.NoError(err, "Failed to store claim")
+
+	c, err = store.Read("foo")
+	is.NoError(err, "Failed to read claim")
+
+	want = map[string]interface{}{
+		"foo-output": true,
+		"bar-output": "baz",
+	}
+	is.Equal(c.Outputs, want, "Wrong outputs on claim")
```
carolynvs: Same here, the parameter order should be reversed.

claim/claimstore_test.go
vdice: Thank you!  Fixed.

claim/claimstore_test.go
vdice: Thank you!  Fixed.

PR #67, opened 2019-07-19
chore(claim.go): remove unused Files field from Claim struct

vdice:
- remove Files field from Claim struct

I couldn't find any use of this field and it doesn't appear to be a part of the [claims spec](https://github.com/deislabs/cnab-spec/blob/master/400-claims.md).


jeremyrickard:
@vdice can you resolve conflicts in this PR?

vdice:
Oops, thanks @jeremyrickard -- updated!

PR #75, opened 2019-07-23
Switch `digest` to `contentDigest` in the json tags

### LINKED ISSUE #59: [spec compliance] Need to replace "digest" with "contentDigest" in Images/InvocationImages
Our `BaseImage` struct has JSON marshalling tags on the Digest field as follow:

```
Digest    string            `json:"digest,omitempty" mapstructure:"digest"`
```

The spec defines [invocation images](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#invocation-images) and [images](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#the-image-map) to have a `contentDigest` field, not a `digest`. We need to update to this.

jeremyrickard:
This PR swaps `contentDigest` in place of `digest` in the json/mapstructure tags for the BaseImage type. Adds a test for marashalling as well.

Fixes #59

### DIFF
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -621,14 +621,17 @@ func TestDigestPresent(t *testing.T) {
 	bundle, err := Unmarshal(bun)
 	require.NoError(t, err, "the bundle should have been valid")
 
-	//invocationImages[] should have exactly 1 value
 	require.Equal(t, 1, len(bundle.InvocationImages), "there should be one invocation image in the bundle")
-	//the invocation image digest should equal sha256:aaaaaaa...
-	assert.Equal(t, "sha256:aaaaaaa...", bundle.InvocationImages[0].Digest)
+	assert.Equal(t,
+		"sha256:decafbad71b4175951f29eb96035604c8cc372c99affa2e6d05cde6e8e20cc9a",
+		bundle.InvocationImages[0].Digest,
+	)
 
-	//images[] should have exactly 1 values
-	_, ok := bundle.Images["my-microservice"]
+	image, ok := bundle.Images["my-microservice"]
 	require.True(t, ok, "there should been an image named my-microservice in the bundle")
-	// the image digest should equal sha256:aaaaaaaaaaaa...
-	assert.Equal(t, "sha256:aaaaaaaaaaaa...", bundle.Images["my-microservice"].Digest)
+	assert.Equal(
+		t,
+		"sha256:beefcacef6c04336a17761db2004813982abe0e87ab727a376c291e09391ea61",
+		image.Digest,
+	)
 }
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -613,3 +613,22 @@ func TestBundle_RoundTrip(t *testing.T) {
 		})
 	}
 }
+
+func TestDigestPresent(t *testing.T) {
+	bun, err := ioutil.ReadFile("../testdata/bundles/digest.json")
+	require.NoError(t, err, "couldn't read test bundle")
+
+	bundle, err := Unmarshal(bun)
+	require.NoError(t, err, "the bundle should have been valid")
+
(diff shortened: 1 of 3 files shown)


testdata/bundles/digest.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "foo",
+  "version": "1.0",
+  "schemaVersion": "99.99",
+  "images": {
+    "my-microservice": {
+      "contentDigest": "sha256:aaaaaaaaaaaa...",
```
glyn: On the basis that open source tests are a form of documentation, I'd prefer it if we used a proper content digest e.g. "sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef" (or a vegetarian alternative).

testdata/bundles/digest.json
```diff
@@ -0,0 +1,67 @@
...
+      "description": "my microservice",
+      "image": "technosophos/microservice:1.2.3"
+    }
+  },
+  "invocationImages": [
+    {
+      "contentDigest": "sha256:aaaaaaa...",
```
glyn: See previous comment about proper content digest, but use a distinct one here.

testdata/bundles/digest.json
```diff
@@ -0,0 +1,67 @@
...
+  },
+  "parameters" : {
+      "serverConfig" : {
+          "definition" : "complexThing",
+          "destination" : {
+              "path": "/cnab/is/go"
+          }
+      }
+  }
+}
```
glyn: nit: add newline to keep github view clean?

bundle/bundle_test.go
```diff
@@ -613,3 +613,22 @@ func TestBundle_RoundTrip(t *testing.T) {
...
+
+	bundle, err := Unmarshal(bun)
+	require.NoError(t, err, "the bundle should have been valid")
+
+	//invocationImages[] should have exactly 1 value
+	require.Equal(t, 1, len(bundle.InvocationImages), "there should be one invocation image in the bundle")
+	//the invocation image digest should equal sha256:aaaaaaa...
+	assert.Equal(t, "sha256:aaaaaaa...", bundle.InvocationImages[0].Digest)
+
+	//images[] should have exactly 1 values
```
glyn:
```suggestion
	//images[] should have exactly 1 value
```

glyn: Interesting that we don't seem to check this. Better to replace the comment with an actual check.

bundle/bundle_test.go
```diff
@@ -613,3 +613,22 @@ func TestBundle_RoundTrip(t *testing.T) {
...
+	_, ok := bundle.Images["my-microservice"]
```
glyn: Why not set the value into a variable here rather than ignoring it. Then the variable can be used in the later check of the digest.

bundle/bundle_test.go
```diff
@@ -613,3 +613,22 @@ func TestBundle_RoundTrip(t *testing.T) {
...
+	require.True(t, ok, "there should been an image named my-microservice in the bundle")
+	// the image digest should equal sha256:aaaaaaaaaaaa...
```
glyn: nit: best to delete comments which merely repeat what the code says.

testdata/bundles/digest.json
jeremyrickard: Yep, good idea. I took this directly from one of the examples in the spec examples directory.

bundle/bundle_test.go
jeremyrickard: removed the comments.

bundle/bundle_test.go
jeremyrickard: Updated.

bundle/bundle_test.go
jeremyrickard: Removed the comments.

PR #79, opened 2019-07-24
Refactor required parameters

### LINKED ISSUE #68: [spec compliance] Update parameters to reflect cnab-spec#227
(https://github.com/deislabs/cnab-spec/pull/227) in https://github.com/deislabs/cnab-spec introduced some changes to how parameters are defined, so we need to update cnab-go to reflect that.

jlegrone:
Relates to #68


bundle/parameters.go
```diff
@@ -1,14 +1,10 @@
...
-
-// ParameterDefinition defines a single parameter for a CNAB bundle
-type ParameterDefinition struct {
+// Parameter defines a single parameter for a CNAB bundle
+type Parameter struct {
 	Definition  string    `json:"definition" mapstructure:"definition"`
 	ApplyTo     []string  `json:"applyTo,omitempty" mapstructure:"applyTo,omitempty"`
 	Description string    `json:"description,omitempty" mapstructure:"description"`
 	Destination *Location `json:"destination,omitemtpty" mapstructure:"destination"`
+	Required    bool
```
vdice: Looks like this needs a corresponding json identifier still: `json:"required" mapstructure:"required"`

jcsirot:
Since `Parameters` is now a map of `Parameter` I think that this check is not required anymore https://github.com/deislabs/cnab-go/blob/22ed003cf91e38449b08dfbf12ba1f5dc0f83344/bundle/bundle.go#L145

bundle/parameters_test.go
```diff
@@ -39,16 +37,14 @@ func TestCanReadParameterDefinition(t *testing.T) {
...
-				"test": {
-					"definition": "%s",
-					"destination": {
-						"env": "%s",
-						"path": "%s"
-					},
-					"description": "%s",
-					"applyTo": [ "%s", "%s" ]
-				}
+			"test": {
```
radu-matei: Could you also add a small check for the required property here?

bundle/bundle.go
```diff
@@ -97,7 +97,7 @@ type InvocationImage struct {
 	BaseImage `mapstructure:",squash"`
 }
 
-// Map that stores the relocated images
+// ImageRelocationMap stores the relocated images
```
radu-matei: Nice catch, thanks!

Issue #73, opened 2019-07-23
[spec compliance] Implement Outputs in Kubernetes Driver


jeremyrickard:
Now that we have mostly added all the relevant output definitions and claims elements, we should complete the picture and implement output handling in the Kubernetes driver. The Kubernetes driver currently handles placing parameter and credentials into files for the invocation image, so we should extend the driver to handle obtaining them from the invocation image, maybe by adding a volume to store the outputs and reading from that? 

I think we should leave _what happens_ to them after they are read from the invocation image up to the end tooling, but we should make the driver able to fetch them.

astrieanna:
We summarized our thoughts on implementation options for the k8s driver here: https://github.com/deislabs/duffle/issues/747#issuecomment-512048160

PR #82, opened 2019-07-25
Add Outputs support to Docker Driver

astrieanna:
In pursuit of https://github.com/deislabs/duffle/issues/747
Fixes https://github.com/deislabs/cnab-go/issues/14 and fixes https://github.com/deislabs/cnab-go/issues/72

This PR:
- In action, if the driver gives you output files, put them in the claim.
- When adding output files to the claim, check that they match the definitions.
- Add Output support in Docker driver, and integration tests for same.
- Make K8s and Command drivers compile, but do not implement support.

cc @youreddy @jeremyrickard

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -5,9 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"strings"
 
 	"github.com/deislabs/cnab-go/bundle"
+	"github.com/deislabs/cnab-go/bundle/definition"
 	"github.com/deislabs/cnab-go/claim"
 	"github.com/deislabs/cnab-go/credentials"
 	"github.com/deislabs/cnab-go/driver"
@@ -27,6 +29,129 @@ const stateful = false
 type Action interface {
 	// Run an action, and record the status in the given claim
 	Run(*claim.Claim, credentials.Set, io.Writer) error
+}
+
+func golangTypeToJSONType(value interface{}) (string, error) {
+	switch v := value.(type) {
+	case nil:
+		return "null", nil
+	case bool:
+		return "boolean", nil
+	case float64:
+		// All numeric values are parsed by JSON into float64s. When a value could be an integer, it could also be a number, so give the more specific answer.
+		if math.Trunc(v) == v {
+			return "integer", nil
+		}
+		return "number", nil
+	case string:
+		return "string", nil
+	case map[string]interface{}:
+		return "object", nil
+	case []interface{}:
+		return "array", nil
+	default:
+		return fmt.Sprintf("%T", value), fmt.Errorf("unsupported type: %T", value)
+	}
+}
+
+// allowedTypes takes an output Schema and returns a map of the allowed types (to true)
+// or an error (if reading the allowed types from the schema failed).
+func allowedTypes(outputSchema definition.Schema) (map[string]bool, error) {
+	var outputTypes []string
+	mapOutputTypes := map[string]bool{}
+
+	// Get list of one or more allowed types for this output
+	outputType, ok, err1 := outputSchema.GetType()
+	if !ok { // there are multiple types
+		var err2 error
+		outputTypes, ok, err2 = outputSchema.GetTypes()
+		if !ok {
+			return mapOutputTypes, fmt.Errorf("Getting a single type errored with %q and getting multiple types errored with %q", err1, err2)
+		}
+	} else {
+		outputTypes = []string{outputType}
+	}
+
+	// Turn allowed outputs into map for easier membership checking
+	for _, v := range outputTypes {
+		mapOutputTypes[v] = true
+	}
+
+	// All integers make acceptable numbers, and our helper function provides the most specific type.
+	if mapOutputTypes["number"] {
+		mapOutputTypes["integer"] = true
+	}
+
+	return mapOutputTypes, nil
+}
+
+// keys takes a map and returns the keys joined into a comma-separate string.
+func keys(stringMap map[string]bool) string {
+	var keys []string
+	for k := range stringMap {
+		keys = append(keys, k)
+	}
+	return strings.Join(keys, ",")
+}
+
+// isTypeOK uses the content and allowedTypes arguments to make sure the content of an output file matches one of the allowed types.
+// The other arguments (name and allowedTypesList) are used when assembling error messages.
+func isTypeOk(name, content string, allowedTypes map[string]bool) error {
+	if !allowedTypes["string"] { // String output types are always passed through as the escape hatch for non-JSON bundle outputs.
+		var value interface{}
+		if err := json.Unmarshal([]byte(content), &value); err != nil {
+			return fmt.Errorf("failed to parse %q: %s", name, err)
+		}
+
+		v, err := golangTypeToJSONType(value)
+		if err != nil {
+			return fmt.Errorf("%q is not a known JSON type. Expected %q to be one of: %s", name, v, keys(allowedTypes))
+		}
+		if !allowedTypes[v] {
+			return fmt.Errorf("%q is not any of the expected types (%s) because it is %q", name, keys(allowedTypes), v)
+		}
+	}
+	return nil
+}
+
+func setOutputsOnClaim(claim *claim.Claim, outputs map[string]string) error {
+	var outputErrors []error
+	claim.Outputs = map[string]interface{}{}
+
+	if claim.Bundle.Outputs == nil {
+		return nil
+	}
+
+	for outputName, v := range claim.Bundle.Outputs.Fields {
+		name := v.Definition
+		if name == "" {
+			return fmt.Errorf("invalid bundle: no definition set for output %q", outputName)
+		}
+
+		outputSchema := claim.Bundle.Definitions[name]
+		if outputSchema == nil {
+			return fmt.Errorf("invalid bundle: output %q references definition %q, which was not found", outputName, name)
+		}
+		outputTypes, err := allowedTypes(*outputSchema)
+		if err != nil {
+			return err
+		}
+
+		content := outputs[v.Path]
+		if content != "" {
+			err := isTypeOk(outputName, content, outputTypes)
+			if err != nil {
+				outputErrors = append(outputErrors, err)
+			}
+			claim.Outputs[outputName] = outputs[v.Path]
+		}
+	}
+
+	if len(outputErrors) > 0 {
+		return fmt.Errorf("error: %s", outputErrors)
+	}
+
+	return nil
 }
 
 func selectInvocationImage(d driver.Driver, c *claim.Claim) (bundle.InvocationImage, error) {
@@ -96,6 +221,13 @@ func opFromClaim(action string, stateless bool, c *claim.Claim, ii bundle.Invoca
 	env["CNAB_BUNDLE_NAME"] = c.Bundle.Name
 	env["CNAB_BUNDLE_VERSION"] = c.Bundle.Version
 
+	var outputs []string
+	if c.Bundle.Outputs != nil {
+		for _, v := range c.Bundle.Outputs.Fields {
+			outputs = append(outputs, v.Path)
+		}
+	}
+
 	return &driver.Operation{
 		Action:       action,
 		Installation: c.Name,
@@ -105,6 +237,7 @@ func opFromClaim(action string, stateless bool, c *claim.Claim, ii bundle.Invoca
 		Revision:     c.Revision,
 		Environment:  env,
 		Files:        files,
+		Outputs:      outputs,
 		Out:          w,
 	}, nil
 }
--- a/action/install_test.go
+++ b/action/install_test.go
@@ -1,9 +1,9 @@
 package action
 
 import (
+	"errors"
 	"io/ioutil"
 	"testing"
-	"time"
 
 	"github.com/deislabs/cnab-go/claim"
 	"github.com/deislabs/cnab-go/driver"
@@ -17,21 +17,66 @@ var _ Action = &Install{}
 func TestInstall_Run(t *testing.T) {
 	out := ioutil.Discard
 
-	c := &claim.Claim{
-		Created:    time.Time{},
-		Modified:   time.Time{},
-		Name:       "name",
-		Revision:   "revision",
-		Bundle:     mockBundle(),
-		Parameters: map[string]interface{}{},
-	}
+	t.Run("happy-path", func(t *testing.T) {
+		c := newClaim()
+		inst := &Install{Driver: &mockDriver{
+			shouldHandle: true,
+			Result: driver.OperationResult{
+				Outputs: map[string]string{
+					"/tmp/some/path": "SOME CONTENT",
+				},
+			},
+			Error: nil,
+		}}
+		assert.NoError(t, inst.Run(c, mockSet, out))
+		assert.Equal(t, claim.StatusSuccess, c.Result.Status)
+		assert.Equal(t, claim.ActionInstall, c.Result.Action)
+		assert.Equal(t, map[string]interface{}{"some-output": "SOME CONTENT"}, c.Outputs)
+	})
 
-	inst := &Install{Driver: &driver.DebugDriver{}}
-	assert.NoError(t, inst.Run(c, mockSet, out))
+	t.Run("when the bundle has no outputs", func(t *testing.T) {
+		c := newClaim()
+		c.Bundle.Outputs = nil
+		inst := &Install{
+			Driver: &mockDriver{
+				shouldHandle: true,
+				Result:       driver.OperationResult{},
+				Error:        nil,
+			},
+		}
+		assert.NoError(t, inst.Run(c, mockSet, out))
+		assert.Equal(t, claim.StatusSuccess, c.Result.Status)
+		assert.Equal(t, claim.ActionInstall, c.Result.Action)
+		assert.Empty(t, c.Outputs)
+	})
 
-	inst = &Install{Driver: &mockFailingDriver{}}
-	assert.Error(t, inst.Run(c, mockSet, out))
+	t.Run("error case: driver can't handle image", func(t *testing.T) {
+		c := newClaim()
+		inst := &Install{
+			Driver: &mockDriver{
+				shouldHandle: false,
+				Error:        errors.New("I always fail"),
+			},
+		}
+		assert.Error(t, inst.Run(c, mockSet, out))
+	})
 
-	inst = &Install{Driver: &mockFailingDriver{shouldHandle: true}}
-	assert.Error(t, inst.Run(c, mockSet, out))
+	t.Run("error case: driver returns error", func(t *testing.T) {
+		c := newClaim()
+		inst := &Install{
+			Driver: &mockDriver{
+				shouldHandle: true,
+				Result: driver.OperationResult{
+					Outputs: map[string]string{
+						"/tmp/some/path": "SOME CONTENT",
+					},
+				},
+				Error: errors.New("I always fail"),
+			},
+		}
+		assert.Error(t, inst.Run(c, mockSet, out))
+		assert.Equal(t, claim.StatusFailure, c.Result.Status)
+		assert.Equal(t, claim.ActionInstall, c.Result.Action)
+		assert.Equal(t, map[string]interface{}{"some-output": "SOME CONTENT"}, c.Outputs)
+	})
 }
--- a/action/run_custom_test.go
+++ b/action/run_custom_test.go
@@ -1,9 +1,9 @@
 package action
 
 import (
+	"errors"
 	"io/ioutil"
 	"testing"
-	"time"
 
 	"github.com/deislabs/cnab-go/claim"
 	"github.com/deislabs/cnab-go/driver"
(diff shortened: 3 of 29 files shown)


jeremyrickard:
🎊 thanks for the hard work @astrieanna @youreddy

driver/docker/docker_integration_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/deislabs/cnab-go/driver"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDriver_Run(t *testing.T) {
+	op := &driver.Operation{
+		Installation: "example",
+		Action:       "install",
+		Image:        "pvtlmc/example-outputs@sha256:ae3c1c719d87e981ecfdf84862ff389ce6a1281c99372f45d158f52e3d5be3dc",
```
silvin-lubecki: Could we build this bundle locally? I'm a little bit worried to run test against an unknown image pushed on a registry. If the image is accidentally deleted the test will fail and we won't be able to re-push the exact same image.
And most of all I think the whole sources of this integration test acts also as an example/documentation, and then without the `cnab/run` sources and/or the `Dockerfile` to produce the image, we are missing something.

driver/docker/docker_integration_test.go
```diff
@@ -0,0 +1,36 @@
...
+		Outputs:      []string{"/cnab/app/outputs/output1", "/cnab/app/outputs/output2"},
+	}
+
+	var output bytes.Buffer
+	op.Out = &output
+	op.Environment = map[string]string{}
```
silvin-lubecki: nit:
```go
op.Environment = map[string]string{
    "CNAB_ACTION":            op.Action,
    "CNAB_INSTALLATION_NAME": op.Installation,
}
```

driver/driver.go
```diff
@@ -33,6 +33,8 @@ type Operation struct {
 	Environment map[string]string `json:"environment"`
 	// Files contains files that should be injected into the invocation image.
 	Files map[string]string `json:"files"`
+	// Outputs
```
silvin-lubecki: Could this comment be more precise? What is intended in this field, paths?

driver/docker/docker.go
```diff
@@ -153,11 +158,41 @@ func (d *Driver) exec(op *driver.Operation) error {
...
+		return driver.OperationResult{}, err
+	}
+	defer os.RemoveAll(outputsFolder)
+
+	// On Mac, the default temp folder (set by the $TMPDIR env variable) is in /var.
+	// Docker for Mac (by default) can access these paths at /private/var but not /var.
+	// To make things work smoothly, we'll adjust the path when no config is set:
+	if runtime.GOOS == "darwin" {
+		if d.config["OUTPUTS_MOUNT_PATH"] == "" && strings.HasPrefix(outputsFolder, "/var") {
+			outputsFolder = "/private" + outputsFolder
```
silvin-lubecki: nit: use `filepath.Join` instead?

driver/docker/docker.go
```diff
@@ -153,11 +158,41 @@ func (d *Driver) exec(op *driver.Operation) error {
...
 	}
 
-	hostCfg := &container.HostConfig{AutoRemove: true}
+	outputsFolder, err := ioutil.TempDir(d.config["OUTPUTS_MOUNT_PATH"], "outputs")
+	if err != nil {
+		return driver.OperationResult{}, err
+	}
+	defer os.RemoveAll(outputsFolder)
+
+	// On Mac, the default temp folder (set by the $TMPDIR env variable) is in /var.
```
silvin-lubecki: Thanks for spotting that, just added this feedback to the Docker Desktop Team, should be shortly fixed 👍

driver/docker/docker.go
```diff
@@ -153,11 +158,41 @@ func (d *Driver) exec(op *driver.Operation) error {
...
+			outputsFolder = "/private" + outputsFolder
+		}
+	}
+
+	err = os.Chmod(outputsFolder, 0777)
+	if err != nil {
+		return driver.OperationResult{}, err
+	}
+
+	hostCfg := &container.HostConfig{
```
silvin-lubecki: I wonder if we should use bind mounts, as it forbids executing the CNAB remotely 🤔 
I guess we should instead copy the output directory, symmetrically to the [inputs](https://github.com/deislabs/cnab-go/blob/master/driver/docker/docker.go#L178) which are tared then copied to the container (and not binded).

action/uninstall_test.go
```diff
@@ -17,42 +17,68 @@ var _ Action = &Uninstall{}
...
-	assert.Error(t, uninst.Run(c, mockSet, out))
-	if c.Result.Message == "" {
-		t.Error("Expected error message in claim result message")
-	}
-
-	if c.Result.Action != claim.ActionUninstall {
-		t.Errorf("Expected claim result action to be %v, got %v", claim.ActionUninstall, c.Result.Action)
-	}
+	// error case: driver doesn't handle image
+	c = newClaim()
```
silvin-lubecki: I think each `c = newClaim()` should be split and put into a golang sub test:
```go
t.Run("happy-path", func(t *testing.T){
...
}
```

action/run_custom_test.go
```diff
@@ -17,35 +17,82 @@ var _ Action = &RunCustom{}
 
 func TestRunCustom(t *testing.T) {
 	out := ioutil.Discard
-	is := assert.New(t)
 
+	// happy path
```
silvin-lubecki: I think each `c = newClaim()` should be split and put into a golang sub test:
```go
t.Run("happy-path", func(t *testing.T){
...
}
```

action/install_test.go
```diff
@@ -17,21 +17,62 @@ var _ Action = &Install{}
...
 
-	c := &claim.Claim{
-		Created:    time.Time{},
-		Modified:   time.Time{},
-		Name:       "name",
-		Revision:   "revision",
-		Bundle:     mockBundle(),
-		Parameters: map[string]interface{}{},
-	}
+	// happy path
```
silvin-lubecki: I think each `c = newClaim()` should be split and put into a golang sub test:
```go
t.Run("happy-path", func(t *testing.T){
...
}
```

action/action.go
```diff
@@ -98,6 +183,13 @@ func opFromClaim(action string, stateless bool, c *claim.Claim, ii bundle.Invoca
 	env["CNAB_BUNDLE_NAME"] = c.Bundle.Name
 	env["CNAB_BUNDLE_VERSION"] = c.Bundle.Version
 
+	var outputs []string
+	if c.Bundle.Outputs != nil {
+		for k := range c.Bundle.Outputs.Fields {
```
silvin-lubecki: nit:
```go
for k, v := range c.Bundle.Outputs.Fields {
    outputs = append(outputs, v.Path)
}
```

action/action.go
```diff
@@ -29,6 +30,90 @@ type Action interface {
...
+		return "object", nil
+	case []interface{}:
+		return "array", nil
+	default:
+		return fmt.Sprintf("%T", value), fmt.Errorf("unsupported type: %T", value)
+	}
+}
+
+func setOutputsOnClaim(claim *claim.Claim, outputs map[string]string) error {
+	outputErrors := []string{}
```
silvin-lubecki: nit: `var outputErrors []string`

action/action.go
```diff
@@ -29,6 +30,90 @@ type Action interface {
...
+	claim.Outputs = map[string]interface{}{}
+
+	if claim.Bundle.Outputs != nil {
```
silvin-lubecki: nit: Early return is easier to read:
```go
if claim.Bundle.Outputs == nil{
    return nil
}
```

action/action.go
```diff
@@ -29,6 +30,90 @@ type Action interface {
...
+		for outputName, v := range claim.Bundle.Outputs.Fields {
+			name := claim.Bundle.Outputs.Fields[outputName].Definition
```
silvin-lubecki: nit:
```go
name := v.Definition
```

action/action.go
```diff
@@ -29,6 +30,90 @@ type Action interface {
...
+			outputSchema := claim.Bundle.Definitions[name]
+
+			var outputTypes []string
+			outputType, ok, _ := outputSchema.GetType()
+			if !ok { // there are multiple types
+				var err error
+				outputTypes, ok, err = outputSchema.GetTypes()
+				if !ok {
+					panic(err)
```
silvin-lubecki: KEEP CALM AND DON'T PANIC 😸 
I think you can early return here with an error?

action/action.go
```diff
@@ -29,6 +30,90 @@ type Action interface {
...
+func setOutputsOnClaim(claim *claim.Claim, outputs map[string]string) error {
+	outputErrors := []string{}
+	claim.Outputs = map[string]interface{}{}
+
+	if claim.Bundle.Outputs != nil {
+		for outputName, v := range claim.Bundle.Outputs.Fields {
+			name := claim.Bundle.Outputs.Fields[outputName].Definition
+			outputSchema := claim.Bundle.Definitions[name]
+
+			var outputTypes []string
```
silvin-lubecki: nit: Can you extract this `outputTypes` and `mapOutputTypes` resolution in a sub function for better readability?

action/action.go
```diff
@@ -29,6 +30,90 @@ type Action interface {
...
+			} else {
+				outputTypes = []string{outputType}
+			}
+
+			mapOutputTypes := map[string]bool{}
+			for _, thing := range outputTypes {
+				mapOutputTypes[thing] = true
+			}
+
+			content := outputs[v.Path]
```
silvin-lubecki: nit: Can you extract this type resolution in a sub-function?

driver/docker/docker.go
youreddy: The reason we chose to mount the directory was to ensure that the files would be available when the container stopped running. We've tried prototyping a version with `CopyFromContainer` subbed into the `fetchOutputs` implementation, but because the container has already stopped by that point, docker can't find the container to create the archive. Perhaps there's a way to restructure the control flow for [executing the container](https://github.com/deislabs/cnab-go/pull/82/files#diff-f7d55a687ffed0311363529637ca9749L221), but I don't think there's a point where you could know the outputs were ready but the container was also still available to read from.

If the driver needs to also support remote Docker servers, I think that's a follow up change someone could make after this PR where the code can use `volumes` instead and the user provides docker config to indicate the driver is not local.

driver/docker/docker_integration_test.go
youreddy: We ended up putting the bundle in our [duffle PR 815](https://github.com/deislabs/duffle/pull/815) under `examples/outputs` since thats where the other test bundles live.

driver/docker/docker.go
silvin-lubecki: We can also:
-  remove the `autoremove` [here](https://github.com/deislabs/cnab-go/blob/master/driver/docker/docker.go#L156)
- defer the container removal
- wait for an `exited` status with the `WaitConditionNotRunning` condition [here](https://github.com/deislabs/cnab-go/blob/master/driver/docker/docker.go#L221)
- then use [CopyFromContainer](https://github.com/docker/cli/blob/f95ca8e1ba6c22c9abcdbf65e8dcc39c53958bba/vendor/github.com/docker/docker/client/interface.go#L76) on the `output` directory?

I guess this should work either locally or remotely ?

WDYT? Am I missing something?

driver/docker/docker_integration_test.go
silvin-lubecki: Ok just checked that, but from a strict dependency point of view, I think cnab-go shouldn't depend from duffle, even indirectly like here. Is it possible to move the `example/outputs` here in cnab-go and reverse the dependency?

youreddy:
@silvin-lubecki we just updated the PR branch and incorporated the suggestions you made. Let us know if you have any more feedback!

driver/docker/docker_integration_test.go
youreddy: That makes sense to me, I can make that update to both PRs

driver/docker/docker.go
astrieanna: Thank you for clarifying what you meant! This does work. :)

astrieanna:
@glyn do you want to take a look at this side of the PR?

jeremyrickard:
Looks like the tests are failing on the linter:


```
action/action.go:92:10: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
		} else {
		       ^
Makefile:25: recipe for target 'lint' failed
```

driver/docker/docker_integration_test.go
```diff
@@ -0,0 +1,37 @@
...
+	var output bytes.Buffer
+	op.Out = &output
+	op.Environment = map[string]string{
+		"CNAB_ACTION":            op.Action,
+		"CNAB_INSTALLATION_NAME": op.Installation,
+	}
+
+	docker := &Driver{}
+	docker.SetContainerOut(op.Out) // Docker driver writes container stdout to driver.containerOut.
+	opResult, err := docker.Run(op)
```
radu-matei: Since this tries to connect to a Docker daemon, we should also update CI to handle this scenario:

```
--- FAIL: TestDriver_Run (0.00s)
    docker_integration_test.go:30: 
        	Error Trace:	docker_integration_test.go:30
        	Error:      	Received unexpected error:
        	            	cannot create container: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
        	Test:       	TestDriver_Run
```

This is why CI is currently failing.

driver/docker/docker.go
```diff
@@ -218,25 +221,71 @@ func (d *Driver) exec(op *driver.Operation) error {
...
+	for err == nil {
+		// skip directories because we're gathering file contents
+		if header.FileInfo().IsDir() {
+			header, err = tarReader.Next()
+			continue
+		}
+
+		var contents []byte
+		// CopyFromContainer strips prefix above outputs directory.
+		pathInContainer := filepath.Join("/cnab/app", header.Name)
```
glyn: I'm wondering about the behaviour in Windows. I think we'd end up with paths like "/cnab/app\x\y\z" which seems odd. Would it be cleaner to use paths as they were in the container? If so, just use `path.Join`. (Also, note that such paths are valid Windows paths FWIW.)

driver/docker/docker_integration_test.go
glyn: Yeah, and this PR should probably also update the README to make it clear that you need a docker daemon in order to run the tests.

glyn:
> In pursuit of [deislabs/duffle#747](https://github.com/deislabs/duffle/issues/747)
> 
> This PR:
> 
> * In action, if the driver gives you output files, put them in the claim.
> * When adding output files to the claim, check that they match the definitions.
> * Add Output support in Docker driver, and integration tests for same.
> * Make K8s and Command drivers compile, but do not implement support.

Why not? This PR will break those drivers. But at least could you raise a couple of issues to cover this and maybe reference them from TODOs in the code? We would then need to implement those issues before bumping duffle to use this PR.

Actually, I think it would be much better to make this PR cover all the drivers in case other issues come out of the woodwork during implementation.

driver/docker/docker.go
youreddy: We gave this a try in the playground and couldn't find a difference between the two https://play.golang.org/p/tQVQNw1YJZR. Is there something we're missing?

action/action.go
```diff
@@ -29,6 +31,122 @@ type Action interface {
...
+		return "array", nil
+	default:
+		return fmt.Sprintf("%T", value), fmt.Errorf("unsupported type: %T", value)
+	}
+}
+
+// allowedTypes takes an output Schema and returns a map of the allowed types (to true)
+// and a string that is a list of all the allowed types (for error messagse)
+// or an error (if reading the allowed types from the schema failed).
+func allowedTypes(outputSchema definition.Schema) (map[string]bool, string, error) {
```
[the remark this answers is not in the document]
astrieanna: What makes the `map[string]struct{}` representation idiomatic?

It looks like [_Effective Go_ approves of `map[string]bool`](https://golang.org/doc/effective_go.html#maps).

action/action.go
```diff
@@ -29,6 +31,122 @@ type Action interface {
...
+		mapOutputTypes[thing] = true
+	}
+
+	return mapOutputTypes, strings.Join(outputTypes, ", "), nil
+}
+
+// isTypeOK uses the content and allowedTypes arguments to make sure the content of an output file matches one of the allowed types.
+// The other arguments (name and allowedTypesList) are used when assembling error messages.
+func isTypeOk(name, content string, allowedTypes map[string]bool, allowedTypesList string) error {
+	if !allowedTypes["string"] { // String output types are always passed through as the escape hatch for non-JSON bundle outputs.
```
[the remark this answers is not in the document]
astrieanna: Yes, if you give an output a definition with type string, then any file will meet the requirement. The idea is that many output files will not be JSON (since most software outputs non-JSON files -- e.g. YAML files, logs, etc). It would be weird to ask bundle authors to insert quotes at the beginning and end of random files that they want to pass back to users just to make them look like strings for JSON.

This should be made more specific in the CNAB spec.

action/install_test.go
```diff
@@ -17,21 +17,66 @@ var _ Action = &Install{}
...
 
-	c := &claim.Claim{
-		Created:    time.Time{},
-		Modified:   time.Time{},
-		Name:       "name",
-		Revision:   "revision",
-		Bundle:     mockBundle(),
-		Parameters: map[string]interface{}{},
-	}
+	t.Run("happy-path", func(t *testing.T) {
```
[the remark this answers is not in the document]
astrieanna: This was requested by @silvin-lubecki . 
https://github.com/deislabs/cnab-go/pull/82#discussion_r307769623

astrieanna:
> > In pursuit of [deislabs/duffle#747](https://github.com/deislabs/duffle/issues/747)
> > * Make K8s and Command drivers compile, but do not implement support.
> 
> Why not? This PR will break those drivers. But at least could you raise a couple of issues to cover this and maybe reference them from TODOs in the code? We would then need to implement those issues before bumping duffle to use this PR.
> 
> Actually, I think it would be much better to make this PR cover all the drivers in case other issues come out of the woodwork during implementation.

The linked issue includes a summary of why adding support to the k8s driver was not implemented in this PR. There is already an issue for implementing support in the k8s driver. https://github.com/deislabs/cnab-go/issues/73

The Command driver execs a process to run the invocation image. It is not clear what the expected interface for returning outputs would be for the executable the driver is calling. I have added an issue for implementing support in the command driver: https://github.com/deislabs/cnab-go/issues/84

This PR touches a lot of code, which creates a greater risk for merge conflicts. It implements the outputs in one of the three drivers, which allows for testing the rest of the code changes. I would prefer to merge this PR sooner, while the community discusses options for implementing support in the other drivers.

driver/docker/docker_integration_test.go
```diff
@@ -0,0 +1,37 @@
+package docker
```
radu-matei: missing `// +build integration` at the beginning of the file?

jeremyrickard:
Yay, all the checks pass now. Thanks @youreddy and @astrieanna! 

@glyn and @silvin-lubecki please re-review when you have a moment!

radu-matei:
That being said, currently the integration tests are not currently run in CI.

jeremyrickard:
@radu-matei we should tackle that with a follow up issue IMO.

radu-matei:
I'm ok with that.

jeremyrickard:
follow up referenced above.

action/action.go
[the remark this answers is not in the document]
glyn: It's just common practice because `struct{}` occupies 0 bytes unlike `bool`. Also, the `map[string]bool` representation isn't ideal because there are three possibilities for any given string (absent, present and mapping to `true`, present and mapping to `false`) which isn't what you really want for a set.

I'm disappointed that Effective Go hasn't picked up on this.

action/install_test.go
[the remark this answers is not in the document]
glyn: Ok, I can live with this style, thanks.

driver/docker/docker.go
glyn: The playground is running on *IX and so `filepath.Join` and `path.Join` give the same results. On Windows, the results would be different. (BTW if `header.Name` can have embedded back-slashes, these should be converted to forward slashes.)

If `header.Name` will have embedded back-slashes on Windows, another way to make `pathInContainer` consistent would be to use:
```
pathInContainer := filepath.Join("cnab", "app", header.Name)
```

glyn:
> This PR touches a lot of code, which creates a greater risk for merge conflicts. It implements the outputs in one of the three drivers, which allows for testing the rest of the code changes. I would prefer to merge this PR sooner, while the community discusses options for implementing support in the other drivers.

Fair enough. After merging this PR, we'll need to implement https://github.com/deislabs/cnab-go/issues/73 and https://github.com/deislabs/cnab-go/issues/84 before we can bump cnab-go in duffle.

driver/docker/docker.go
glyn: It would be helpful if someone with a Windows system could try this PR and see how it handles outputs in multiple directories.

youreddy:
> > This PR touches a lot of code, which creates a greater risk for merge conflicts. It implements the outputs in one of the three drivers, which allows for testing the rest of the code changes. I would prefer to merge this PR sooner, while the community discusses options for implementing support in the other drivers.
> 
> Fair enough. After merging this PR, we'll need to implement #73 and #84 before we can bump cnab-go in duffle.

@glyn I'm curious what your specific concerns are for bumping cnab-go before the other drivers are have output support. Are you worried about uncovering issues when implementing outputs in the other drivers or about the specific user experience regression in duffle?

jeremyrickard:
> After merging this PR, we'll need to implement #73 and #84 before we can bump cnab-go in duffle.

I don't think it's _necessary_ to do that. This is _adding_ functionality and not breaking the experience (i.e. outputs didn't exist before), seems like worthy of a a minor release for docker output functionality, then another minor release in duffle (and cnab-go) for k8s and command driver outputs.

driver/docker/docker.go
youreddy: Presumably some of our friends at Microsoft could help us out with that :)
@vdice @jeremyrickard @radu-matei

action/action.go
```diff
@@ -29,6 +31,129 @@ type Action interface {
...
+func isTypeOk(name, content string, allowedTypes map[string]bool) error {
+	if !allowedTypes["string"] { // String output types are always passed through as the escape hatch for non-JSON bundle outputs.
+		var value interface{}
+		if err := json.Unmarshal([]byte(content), &value); err != nil {
+			return fmt.Errorf("failed to parse %q: %s", name, err)
+		}
+
+		v, err := golangTypeToJSONType(value)
+		if err != nil {
+			return fmt.Errorf("%q is not a known JSON type it is %q; expected one of: %s", name, v, keys(allowedTypes))
```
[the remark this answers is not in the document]
youreddy: good point! reworded to: `...is not a known JSON type. Expected 'something' to be one of...`

action/action.go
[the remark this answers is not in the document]
astrieanna: I'm not worried about the memory allocation here. This is a set of allowed types names; there's a small a constant number of JSON types, so in practice this will be tiny.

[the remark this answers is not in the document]
glyn: I agree. I think the representation issue is more important, but I don't want to argue and it will be easy to switch the types later.

glyn:
> > > This PR touches a lot of code, which creates a greater risk for merge conflicts. It implements the outputs in one of the three drivers, which allows for testing the rest of the code changes. I would prefer to merge this PR sooner, while the community discusses options for implementing support in the other drivers.
> > 
> > 
> > Fair enough. After merging this PR, we'll need to implement #73 and #84 before we can bump cnab-go in duffle.
> 
> @glyn I'm curious what your specific concerns are for bumping cnab-go before the other drivers are have output support. Are you worried about uncovering issues when implementing outputs in the other drivers or about the specific user experience regression in duffle?

Since this PR only claims to get the other drivers to compile, I was concerned that their function might regress. If it leaves them usable but without output support, that's fine. /cc @jeremyrickard

action/action.go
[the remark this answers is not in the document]
glyn: For the record, Dave Cheney also [prefers](https://dave.cheney.net/2014/03/25/the-empty-struct) `map[T]bool` as way of representing a set of `T`. I still prefer `map[T]struct{}`, but it's only a preference.

astrieanna:
@jeremyrickard On slack, it sounded like you might be our Windows tester?

silvin-lubecki:
@astrieanna there's a conflict on `action/action_test.go`, can you rebase this PR? 🦁

astrieanna:
> @astrieanna there's a conflict on `action/action_test.go`, can you rebase this PR? 🦁

I've been holding off on rebasing until we've got a green-light to merge otherwise (to avoid resolving conflicts more than once). Is the rebase needed before testing the paths on windows?

driver/docker/docker.go
```diff
@@ -218,25 +221,71 @@ func (d *Driver) exec(op *driver.Operation) error {
...
+			return opResult, fmt.Errorf("error in container: %v", err)
 		}
 	case s := <-statusc:
 		if s.StatusCode == 0 {
-			return nil
+			return d.fetchOutputs(ctx, resp.ID)
 		}
 		if s.Error != nil {
-			return fmt.Errorf("container exit code: %d, message: %v", s.StatusCode, s.Error.Message)
+			opResult, _ := d.fetchOutputs(ctx, resp.ID)
```
silvin-lubecki: This error is dropped, so the user won't be able to understand why she can't get her logs if the operation fails (which can occur as the container itself failed). I guess we should somehow merge the 2 errors 🤔

driver/docker/docker.go
```diff
@@ -218,25 +221,71 @@ func (d *Driver) exec(op *driver.Operation) error {
...
+			return opResult, fmt.Errorf("container exit code: %d, message: %v", s.StatusCode, s.Error.Message)
 		}
-		return fmt.Errorf("container exit code: %d", s.StatusCode)
+		opResult, _ := d.fetchOutputs(ctx, resp.ID)
```
silvin-lubecki: Same here for the error dropped.

driver/docker/docker.go
silvin-lubecki: I agree with @glyn about `pathInContainer := filepath.Join("cnab", "app", header.Name)` 👍

driver/docker/docker_integration_test.go
silvin-lubecki: I saw you added the bundle.json and the Dockerfile needed to be able to rebuild the bundle, but the test is still depending on an image pushed on the Docker Hub. Do you think we can build the image locally before running the test (or maybe directly inside the test, invoking docker?) ?

driver/docker/docker.go
youreddy: Makes sense, I merge those errors.

silvin-lubecki: @youreddy I see this comment marked as resolved but couldn't see the associated commit 🤔

driver/docker/docker.go
glyn: Based on @jeremyrickard's experiments on Windows and if the variable name `pathInContainer` (where we mean a Linux container) is to be taken literally, then the path separator should be `/` and so it's probably best to use `path.Join` instead of `filepath.Join` along these lines:
```suggestion
		pathInContainer := path.Join("cnab", "app", header.Name)
```

jeremyrickard:
@youreddy @astrieanna  sorry for the delay. I ran your test on my windows machine after fixing it:

```
filePath Join:  \cnab\app\outputs\x\y\z
justPath Join:  /cnab/app/outputs\x\y\z
```

I think that since we're _assuming_ that we're reading files from a linux based path (/cnab/app/output) and we're populating that map with that path, I don't think filepath is appropriate due to the cross platform use case (windows machine -> linux docker image). 


I think the more appropratie thing for now is 

```
pathInContainer := path.Join("cnab", "app", header.Name)
```

We'll probably need to revisit with Windows containers, but lets not support that use case yet.

driver/docker/docker.go
youreddy: @silvin-lubecki sorry about that, @astrieanna might have resolved it before I pushed it up. You should be able to see it now!

youreddy:
I just pushed up the `path.Join` and the error handling changes. I think the only thing left is @silvin-lubecki's request for building the image before or during the docker integration tests, which I'm working on.

driver/docker/docker_integration_test.go
youreddy: @silvin-lubecki just added a build script for this. Let me know what you think!

radu-matei:
I think this needs a rebase, and it should be good to go?

PR #83, opened 2019-07-26
feat(bundle.go): add RequiredExtensions

### LINKED ISSUE #71: [spec compliance] Add `requiredExtensions` to Bundle struct
https://github.com/deislabs/cnab-spec/pull/220 in https://github.com/deislabs/cnab-spec introduced the concept of `requiredExtensions`. We should implement appropriate handling of this within cnab-go by adding it to the Bundle struct.

vdice:
- adds the `RequiredExtensions` string array field to the `Bundle` struct, per the CNAB Spec (added in https://github.com/deislabs/cnab-spec/pull/220)

Closes #71

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -16,20 +16,21 @@ import (
 
 // Bundle is a CNAB metadata document
 type Bundle struct {
-	SchemaVersion    string                 `json:"schemaVersion" mapstructure:"schemaVersion"`
-	Name             string                 `json:"name" mapstructure:"name"`
-	Version          string                 `json:"version" mapstructure:"version"`
-	Description      string                 `json:"description" mapstructure:"description"`
-	Keywords         []string               `json:"keywords,omitempty" mapstructure:"keywords"`
-	Maintainers      []Maintainer           `json:"maintainers,omitempty" mapstructure:"maintainers"`
-	InvocationImages []InvocationImage      `json:"invocationImages" mapstructure:"invocationImages"`
-	Images           map[string]Image       `json:"images,omitempty" mapstructure:"images"`
-	Actions          map[string]Action      `json:"actions,omitempty" mapstructure:"actions"`
-	Parameters       *ParametersDefinition  `json:"parameters,omitempty" mapstructure:"parameters"`
-	Credentials      map[string]Credential  `json:"credentials,omitempty" mapstructure:"credentials"`
-	Outputs          *OutputsDefinition     `json:"outputs,omitempty" mapstructure:"outputs"`
-	Definitions      definition.Definitions `json:"definitions,omitempty" mapstructure:"definitions"`
-	License          string                 `json:"license,omitempty" mapstructure:"license"`
+	SchemaVersion      string                 `json:"schemaVersion" mapstructure:"schemaVersion"`
+	Name               string                 `json:"name" mapstructure:"name"`
+	Version            string                 `json:"version" mapstructure:"version"`
+	Description        string                 `json:"description" mapstructure:"description"`
+	Keywords           []string               `json:"keywords,omitempty" mapstructure:"keywords"`
+	Maintainers        []Maintainer           `json:"maintainers,omitempty" mapstructure:"maintainers"`
+	InvocationImages   []InvocationImage      `json:"invocationImages" mapstructure:"invocationImages"`
+	Images             map[string]Image       `json:"images,omitempty" mapstructure:"images"`
+	Actions            map[string]Action      `json:"actions,omitempty" mapstructure:"actions"`
+	Parameters         *ParametersDefinition  `json:"parameters,omitempty" mapstructure:"parameters"`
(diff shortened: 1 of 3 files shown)


bundle/bundle.go
```diff
@@ -191,6 +192,12 @@ func (b Bundle) Validate() error {
 		return errors.New("'latest' is not a valid bundle version")
 	}
 
+	for _, requiredExtension := range b.RequiredExtensions {
```
vdice: I'm actually slightly unsure if this logic matches up to:
> The requiredExtensions array SHOULD contain the EXTENSION NAME defined in the custom object for each extension that is required.

as mentioned in https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#required-extensions

I'm wondering how a required extension would be utilized successfully if not defined in the `Custom` section?

bundle/bundle_test.go
```diff
@@ -375,6 +399,10 @@ func TestReadCustomExtensions(t *testing.T) {
 	}
 	assert.Equal(t, true, backupExt["enabled"])
 	assert.Equal(t, "daily", backupExt["frequency"])
+
+	if len(bundle.RequiredExtensions) != 1 {
```
glyn: Probably best to check the value of the required extension too.

bundle/bundle.go
```diff
@@ -191,6 +192,12 @@ func (b Bundle) Validate() error {
...
+		if _, exists := b.Custom[requiredExtension]; !exists {
+			return fmt.Errorf("required extension '%s' must be defined in the Custom section of the bundle", requiredExtension)
```
glyn: nit:
```suggestion
			return fmt.Errorf("required extension '%s' is not defined in the Custom section of the bundle", requiredExtension)
```

bundle/bundle_test.go
```diff
@@ -339,7 +339,31 @@ func TestValidateBundle_RequiresInvocationImage(t *testing.T) {
...
+		Version:          "0.1.0",
+		SchemaVersion:    "99.98",
+		InvocationImages: []InvocationImage{img},
+		RequiredExtensions: []string{
+			"my.custom.extension",
+		},
+	}
+
+	err := b.Validate()
+	is.EqualError(err, "required extension 'my.custom.extension' must be defined in the Custom section of the bundle")
```
glyn: nit (part 2):
```suggestion
	is.EqualError(err, "required extension 'my.custom.extension' is not defined in the Custom section of the bundle")
```

bundle/bundle.go
```diff
@@ -16,19 +16,20 @@ import (
...
+	Keywords           []string               `json:"keywords,omitempty" mapstructure:"keywords"`
+	Maintainers        []Maintainer           `json:"maintainers,omitempty" mapstructure:"maintainers"`
+	InvocationImages   []InvocationImage      `json:"invocationImages" mapstructure:"invocationImages"`
+	Images             map[string]Image       `json:"images,omitempty" mapstructure:"images"`
+	Actions            map[string]Action      `json:"actions,omitempty" mapstructure:"actions"`
+	Parameters         *ParametersDefinition  `json:"parameters,omitempty" mapstructure:"parameters"`
+	Credentials        map[string]Credential  `json:"credentials,omitempty" mapstructure:"credentials"`
+	Outputs            *OutputsDefinition     `json:"outputs,omitempty" mapstructure:"outputs"`
+	Definitions        definition.Definitions `json:"definitions,omitempty" mapstructure:"definitions"`
+	RequiredExtensions []string               `json:"requiredExtensions,omitempty" mapstructure:"requiredExtensions"`
```
glyn: Since this slice can contain duplicate entries, should we add a test to cover that case and nail down what the behaviour should be?

vdice: Ah, good point.  Should we just check that there are no duplicates during validation?  If there are duplicates, throw an error mentioning this?

radu-matei: The slice can also contain items that are not present in the `custom` map, and we should probably validate that as well.

bundle/bundle.go
radu-matei: Yeah, I think this check is enough to validate that the items in `requiredExtensions` are defined in `custom`.

bundle/bundle.go
glyn: > Ah, good point. Should we just check that there are no duplicates during validation? If there are duplicates, throw an error mentioning this?

Yes, that's probably better than issuing a warning and proceeding as that would encourage duplication/sloppiness.

vdice:
> This looks great, a single mention to maybe also test the case when requiredExtensions contains an item that is not present in custom.

Good call; see unit test added -- it should cover scenario mentioned.

> That being said, we should also expose how drivers advertise what custom extensions they can handle, and check before actually executing the action - should it also be part of this PR, or should we merge this sooner to be in sync with the bundle definition?

I would suggest we tackle this in a different ticket/PR.

bundle/bundle.go
vdice: Updated; thanks!

PR #85, opened 2019-07-30
Remove fields from outputs

radu-matei:
https://github.com/deislabs/cnab-spec/issues/234 removes `fields` from outputs, and this PR is issued at the same time with the spec change.

Since #82 adds runtime support for outputs, there is also the option to directly roll this change there - I don't have a strong preference.


jeremyrickard:
Let's do #82 first, then circle back to this. that one has had enough churn and i'd prefer incremental changes.

silvin-lubecki:
@radu-matei can you rebase this PR? 🦁

radu-matei:
@silvin-lubecki - I'd rather wait for #82, then do a rebase.

silvin-lubecki:
Still LGTM 👍

Issue #87, opened 2019-07-30
Add integration tests to CI


jeremyrickard:
PR #82 makes a modification to the Makefile and tests to introduce some integration tests:

```
.PHONY: integration-test
integration-test:
	go test -tags integration ./...
```

Those are not currently run as part of CI, so we need to enhance CI (azure pipelines and brigade) to handle running the integration tests as part of a CI run. Integration requests will require DinD.

PR #89, opened 2019-08-01
[WIP] Brigade integration tests

radu-matei:
addresses #87 

depends on #82

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,10 @@ build:
 test:
 	go test ./...
 
+.PHONY: integration-test
+integration-test:
+	go test -tags integration ./...
+
 .PHONY: lint
 lint:
 	golangci-lint run --config ./golangci.yml
--- a/README.md
+++ b/README.md
@@ -43,10 +43,18 @@ Compile all the code:
 $ make build
 ```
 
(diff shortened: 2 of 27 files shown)


brigade.js
```diff
@@ -4,6 +4,8 @@ const projectOrg = "deislabs";
 const projectName = "cnab-go";
 
 const goImg = "golang:1.11";
+const goDind = "radumatei/golang-dind:1.11-dev"
```
radu-matei: We should also track the Dockerfile for this image somewhere - it starts from DinD and adds Go 1.11.12.

radu-matei:
Update on running `kind` in a Brigade job - there is now a PR on Brigade that allows us to mount the volumes necessary for `kind` to run in a Kubernetes pod.

To enable it for this project, we need to:

- set the custom worker tag for the project
- make sure project allows privileged jobs and host mounts
- actually update the integration test pipeline to first start the `kind` cluster.

I can update the last point, but I'll have to ping @vdice for the first two.

radu-matei:
Removing the "Draft" status to trigger Azure Pipelines builds.

PR #92, opened 2019-08-07
feat(bundle_test.go): add yaml marshal test for bundle

vdice:
~~Adds yaml tags to the `Location` struct for third-party consumers representing these structs in yaml.  See also https://github.com/deislabs/cnab-go/pull/90 for precedence.~~

~~Even though this covers the extent of our needs in Porter as of writing, I'd be willing to add yaml tags to all of the structs, as there may be other uses in the future.  What do we think?~~

**Edit: https://github.com/deislabs/cnab-go/pull/95 covers adding yaml tags to all structs; therefore this PR now just represents some added tests**

Note that the test amendments are largely just extracting the `exampleCred` and `exampleBundle` vars to the top-level such that multiple tests can make use of them.  Also added an additional `Path` field to the `"port"` parameter `Destination` field, for verification in the added yaml test.


bundle/bundle_test.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/deislabs/cnab-go/bundle/definition"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	yaml "gopkg.in/yaml.v2"
```
silvin-lubecki: Nit: I think aliasing this import is no-op, as golang already takes the part before the first point as import name.

PR #94, opened 2019-08-07
feat(*): update Claim per cnab-spec

vdice:
Updates the `Claim` struct to represent the current spec per https://github.com/deislabs/cnab-spec/blob/master/400-claims.md

- removes `RelocationMap` field (and logic/tests) per spec designation of `additionalProperties` as `false`
- adds a `Custom` field  (however, note issue seeking clarification in https://github.com/deislabs/cnab-spec/issues/240)
- updates some fields as optional (`omitempty` on the json tag), per spec
- adds an in-progress schema validation test (mentioning issues encountered in https://github.com/deislabs/cnab-spec/pull/243 and https://github.com/deislabs/cnab-spec/issues/241

Closes https://github.com/deislabs/cnab-go/issues/74


Gopkg.toml
```diff
@@ -19,6 +19,10 @@
   name = "github.com/docker/cli"
   revision = "f95ca8e1ba6c22c9abcdbf65e8dcc39c53958bba"
 
+[[constraint]]
```
silvin-lubecki: Out of curiosity, why do you need this constraint, as it seems it's no-op (no vendoring update or no Gopkg.lock)? 🤔

action/action.go
```diff
@@ -36,11 +36,6 @@ func selectInvocationImage(d driver.Driver, c *claim.Claim) (bundle.InvocationIm
 
 	for _, ii := range c.Bundle.InvocationImages {
 		if d.Handles(ii.ImageType) {
-			if c.RelocationMap != nil {
```
silvin-lubecki: I think here we are removing a feature from cnab-go, pushing it back to runtime developers to "re"-implement this feature on their own.
Maybe instead of just removing it, we could just move it to somewhere else, like maybe in an helper?
This helper would then store and retrieve relocation maps from the custom field.
WDYT?

Gopkg.toml
vdice: Ah, good call; removed.  The package was already included.

action/action.go
vdice: Fair point.  My understanding was indeed that runtime tool authors would need to implement this piece of optional functionality in an additive capacity.  The ticket for this work mentioned the intention for image relocation to be additive:
> Duffle implements this capability, so it could be an additive thing for Duffle. 

Perusing the Duffle code, it appears that it does indeed already have its own logic, utilizing https://github.com/pivotal/image-relocation as a library resource.  It appears to do so outside of the Claim struct, where logic in this repo existed.

If `pivotal/image-relocation` is indeed intended to provide functionality mentioned in https://github.com/deislabs/cnab-spec/blob/master/103-bundle-runtime.md#image-relocation, should we standardize around `pivotal/image-relocation` as the library of choice, instead of keeping it in this repo?

Definitely not familiar enough with this area to make an informed decision.  cc @glyn who might also offer assistance/opinion here.

glyn: https://github.com/pivotal/image-relocation is intended as a general solution to the core problems of image relocation. It is shared by a number of projects including duffle and is not currently tied to CNAB in any way. If there are some general helpers that need to be shared by CNAB runtimes in support of image relocation, it may be acceptable to add some packages (with `cnab` in them, to make it clear they are CNAB-specific).

That said, unless duffle going to add a custom field to claims containing a relocation map, it's going to be hard to get the helpers right as they won't have a "customer" to define the requirements. I wonder if it might be better to wait until the need actually arises.

PR #95, opened 2019-08-08
ref(bundle/*): replace mapstructure tags with yaml tags

vdice:
Represents a possible dual closing of https://github.com/deislabs/cnab-go/issues/91 and adding yaml tags to all of the structs (per recent history of third-party use in https://github.com/deislabs/cnab-go/pull/90 and https://github.com/deislabs/cnab-go/pull/92; 92 could/would be superseded by this PR.)

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -16,25 +16,25 @@ import (
 
 // Bundle is a CNAB metadata document
 type Bundle struct {
-	SchemaVersion      string                 `json:"schemaVersion" mapstructure:"schemaVersion"`
-	Name               string                 `json:"name" mapstructure:"name"`
-	Version            string                 `json:"version" mapstructure:"version"`
-	Description        string                 `json:"description" mapstructure:"description"`
-	Keywords           []string               `json:"keywords,omitempty" mapstructure:"keywords"`
-	Maintainers        []Maintainer           `json:"maintainers,omitempty" mapstructure:"maintainers"`
-	InvocationImages   []InvocationImage      `json:"invocationImages" mapstructure:"invocationImages"`
-	Images             map[string]Image       `json:"images,omitempty" mapstructure:"images"`
-	Actions            map[string]Action      `json:"actions,omitempty" mapstructure:"actions"`
-	Parameters         map[string]Parameter   `json:"parameters,omitempty" mapstructure:"parameters"`
-	Credentials        map[string]Credential  `json:"credentials,omitempty" mapstructure:"credentials"`
-	Outputs            map[string]Output      `json:"outputs,omitempty" mapstructure:"outputs"`
-	Definitions        definition.Definitions `json:"definitions,omitempty" mapstructure:"definitions"`
-	License            string                 `json:"license,omitempty" mapstructure:"license"`
-	RequiredExtensions []string               `json:"requiredExtensions,omitempty" mapstructure:"requiredExtensions"`
+	SchemaVersion      string                 `json:"schemaVersion" yaml:"schemaVersion"`
+	Name               string                 `json:"name" yaml:"name"`
+	Version            string                 `json:"version" yaml:"version"`
+	Description        string                 `json:"description" yaml:"description"`
+	Keywords           []string               `json:"keywords,omitempty" yaml:"keywords,omitempty"`
+	Maintainers        []Maintainer           `json:"maintainers,omitempty" yaml:"maintainers,omitempty"`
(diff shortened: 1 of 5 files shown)


jeremyrickard:
Do we _need_ the YAML? The spec only really supports JSON for the bundle metadata right? 

Is this for downstream use by Porter?

bundle/bundle.go
```diff
@@ -16,25 +16,25 @@ import (
...
-	Definitions        definition.Definitions `json:"definitions,omitempty" mapstructure:"definitions"`
-	License            string                 `json:"license,omitempty" mapstructure:"license"`
-	RequiredExtensions []string               `json:"requiredExtensions,omitempty" mapstructure:"requiredExtensions"`
+	SchemaVersion      string                 `json:"schemaVersion" yaml:"schemaVersion"`
+	Name               string                 `json:"name" yaml:"name"`
+	Version            string                 `json:"version" yaml:"version"`
+	Description        string                 `json:"description" yaml:"description"`
+	Keywords           []string               `json:"keywords,omitempty" yaml:"keywords,omitempty"`
+	Maintainers        []Maintainer           `json:"maintainers,omitempty" yaml:"maintainers,omitempty"`
+	InvocationImages   []InvocationImage      `json:"invocationImages" yaml:"invocationImages,omitempty"`
```
silvin-lubecki: This one should not be `omitempty` I guess.

bundle/bundle.go
```diff
@@ -72,30 +72,30 @@ func (b Bundle) WriteTo(w io.Writer) (int64, error) {
...
+	Size      uint64            `json:"size,omitempty" yaml:"size,omitempty"`
+	Labels    map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
+	MediaType string            `json:"mediaType,omitempty" yaml:"mediaType,omitempty"`
 }
 
 // Image describes a container image in the bundle
 type Image struct {
-	BaseImage   `mapstructure:",squash"`
-	Description string `json:"description" mapstructure:"description"` //TODO: change? see where it's being used? change to description?
+	BaseImage   `yaml:",squash"`
```
silvin-lubecki: I don't think `yaml` understands `squash` annotation 🤔

bundle/bundle.go
```diff
@@ -72,30 +72,30 @@ func (b Bundle) WriteTo(w io.Writer) (int64, error) {
...
+	Description string `json:"description" yaml:"description"` //TODO: change? see where it's being used? change to description?
 }
 
 // InvocationImage contains the image type and location for the installation of a bundle
 type InvocationImage struct {
-	BaseImage `mapstructure:",squash"`
+	BaseImage `yaml:",squash"`
```
silvin-lubecki: Same here

bundle/credentials.go
```diff
@@ -2,7 +2,7 @@ package bundle
 
 // Credential represents the definition of a CNAB credential
 type Credential struct {
-	Location    `mapstructure:",squash"`
-	Description string `json:"description,omitempty" mapstructure:"description"`
-	Required    bool   `json:"required,omitempty" mapstructure:"required"`
+	Location    `yaml:",squash"`
```
silvin-lubecki: Same here

bundle/bundle.go
vdice: oops!  yes thank you, will fix those up to `inline`

vdice:
> Do we need the YAML? The spec only really supports JSON for the bundle metadata right?

It's true, an argument can be made that `yaml` tags shouldn't be in (similar w/ `mapstructure`) b/c cnab-go itself doesn't (un)marshal yaml.  However, indeed, there is one current use of a third-party (Porter) benefiting from having yaml tags on these structs to enabling unmarshaling from yaml into these canonical golang structs.

Definitely open to suggestions on other routes to achieve the same goal of third-party consumers wishing to re-use these canonical golang structs but may be unmarshaling from representations other than JSON.

bundle/bundle.go
vdice: Nice catch! Fixed.

jeremyrickard:
I'm fine with adding the YAML, it makes things more flexible for downstream consumers.

glyn:
If we ever need to revisit the decision to support YAML tags, we might want to try converting from YAML to JSON and then reusing the JSON tags for demarshalling as in this [article](https://sharpend.io/blog/teaching-go-programs-to-love-json-and-yaml/).

PR #96, opened 2019-08-09
Fixes bug where cnab-go was not marshaling params as json

youreddy:
Resolves https://github.com/deislabs/cnab-go/issues/86

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -251,7 +251,22 @@ func injectParameters(action string, c *claim.Claim, env, files map[string]strin
 			}
 			continue
 		}
-		value := fmt.Sprintf("%v", rawval)
+
+		contents, err := json.Marshal(rawval)
+		if err != nil {
+			return err
+		}
+
+		// In order to preserve the exact string value the user provided
+		// we don't marshal string parameters
+		value := string(contents)
+		if value[0] == '"' {
+			value, ok = rawval.(string)
+			if !ok {
+				return fmt.Errorf("failed to parse parameter %q as string", k)
+			}
+		}
+
 		if param.Destination == nil {
 			// env is a CNAB_P_
 			env[fmt.Sprintf("CNAB_P_%s", strings.ToUpper(k))] = value
--- a/action/action_test.go
+++ b/action/action_test.go
@@ -86,28 +86,26 @@ func mockBundle() *bundle.Bundle {
 				Default: "three",
 			},
 			"NullParam": {
-				Type:    "null",
-				Default: true,
+				Type: "null",
 			},
 			"BooleanParam": {
 				Type:    "boolean",
 				Default: true,
 			},
 			"ObjectParam": {
-				Type:    "object",
-				Default: true,
+				Type: "object",
 			},
 			"ArrayParam": {
-				Type:    "array",
-				Default: true,
+				Type: "array",
 			},
 			"NumberParam": {
-				Type:    "number",
-				Default: true,
+				Type: "number",
 			},
 			"IntegerParam": {
-				Type:    "integer",
-				Default: true,
+				Type: "integer",
+			},
+			"StringParam": {
+				Type: "string",
 			},
 			"BooleanAndIntegerParam": {
 				Type: []interface{}{"boolean", "integer"},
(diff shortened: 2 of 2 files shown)


action/action.go
```diff
@@ -251,7 +251,11 @@ func injectParameters(action string, c *claim.Claim, env, files map[string]strin
 			}
 			continue
 		}
-		value := fmt.Sprintf("%v", rawval)
+		contents, err := json.Marshal(rawval)
+		if err != nil {
+			return err
+		}
+		value := strings.Trim(string(contents), "\"")
```
silvin-lubecki: I'm wondering if this `Trim` applied to every contents may alter original content, like a string parameter with double quotes (like `"quoted value"`)?

jeremyrickard: did some quick experiments: 

```
func main() {

	rawval := "[\"first-value\",\"second-value\"]"
	contents, _ := json.Marshal(rawval)
	value := strings.Trim(string(contents), "\"")

	fmt.Printf("Val: %s\n", value)
}
```

That gives `Val: [\"first-value\",\"second-value\"]`

```
func main() {

	rawval := "quoted value"
	contents, _ := json.Marshal(rawval)
	value := strings.Trim(string(contents), "\"")

	fmt.Printf("Val: %s\n", value)
}
```
Gives `Val: quoted value`

```
func main() {

	rawval := "\"escaped quotes\""
	contents, _ := json.Marshal(rawval)
	value := strings.Trim(string(contents), "\"")

	fmt.Printf("Val: %s\n", value)
}
```

Gives: `Val: \"escaped quotes\`

That last one seems like it wouldn't really reflect what the user had passed?

youreddy: Yeah that's not good. I tested this locally as well and rebuilt duffle with and without `strings.Trim`. I came across the same findings as Jeremey where trimming escaped quotes caused unexpected results for the string param. As far as I can tell, removing the `strings.Trim` gives us the desired behavior without causing a problem with escaped quote values. I'm going to update this PR and remove that line.

youreddy: @jeremyrickard @silvin-lubecki it turned out that the simplest thing to do was to keep string values in their raw format. It gives us the desired behavior with strings without having additional logic that tries to figure out when to trim the extra quotes (which felt error prone).

action/action.go
```diff
@@ -251,7 +251,22 @@ func injectParameters(action string, c *claim.Claim, env, files map[string]strin
...
+
+		contents, err := json.Marshal(rawval)
+		if err != nil {
+			return err
+		}
+
+		// In order to preserve the exact string value the user provided
+		// we don't marshal string parameters
+		value := string(contents)
+		if value[0:1] == `"` {
```
silvin-lubecki: nit: `if value[0] == '"'`?

action/action.go
```diff
@@ -251,7 +251,22 @@ func injectParameters(action string, c *claim.Claim, env, files map[string]strin
 			}
 			continue
 		}
-		value := fmt.Sprintf("%v", rawval)
+
+		contents, err := json.Marshal(rawval)
```
silvin-lubecki: I wonder if the following code could be simpler, with something like this:
```go
value, ok = rawval.(string)
if !ok {
    // rawval is not a string, we marshal it as JSON
    contents, err := json.Marshal(rawval)
    if err != nil {
	return err
    }
    value := string(contents)
}
```
WDYT?
Sorry about this back and fourth 😅

astrieanna: The reason we marshal to strings here only to throw the work away was I worried about knowing (for sure) that the golang value would only marshal to a string if its type was already a string. Since this is a library that could be used by CLI implementations other than Duffle, I feel like other marshal-to-string-but-aren't-literally-strings types could end up here and it seems friendlier to give a clear error message rather than ending up with bonus quotes in the values.

silvin-lubecki: Right that makes sense!

astrieanna:
@jeremyrickard Could you review this? It's waiting for a second approval.

astrieanna:
Could we get a second review/approval for this? @radu-matei @vdice

PR #99, opened 2019-08-10
ref(bundle): add ActionApplicable interface, AppliesTo methods to Parameter and Output

vdice:
Add an `ActionApplicable` interface for objects that have the ability to apply to a set of actions (Parameter and Output, currently)

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -223,3 +223,9 @@ func validateDockerish(s string) error {
 	}
 	return nil
 }
+
+// ActionApplicable is an interface for objects that may apply to a subset of actions
+// e.g., Parameters and Outputs
+type ActionApplicable interface {
+	AppliesTo(action string) bool
+}
--- a/bundle/outputs.go
+++ b/bundle/outputs.go
@@ -6,3 +6,17 @@ type Output struct {
 	Description string   `json:"description,omitempty" yaml:"description,omitempty"`
 	Path        string   `json:"path" yaml:"path"`
 }
+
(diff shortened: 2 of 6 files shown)


bundle/bundle.go
```diff
@@ -223,3 +223,9 @@ func validateDockerish(s string) error {
 	}
 	return nil
 }
+
+// ActionApplicable is an interface for objects that may apply to a subset of actions
+// e.g., Parameters and Outputs
+type ActionApplicable interface {
```
silvin-lubecki: I wonder if it's this library's responsibility to provide the ActionApplicable interface. It's mostly a matter of taste I guess, but I think (with golang at least) it should be the library consumer's responsibility to declare and use this interface, if she really needs it.
One hint is that this interface is not used at all in cnab-go.
WDYT?

bundle/outputs.go
```diff
@@ -6,3 +6,17 @@ type Output struct {
 	Description string   `json:"description,omitempty" yaml:"description,omitempty"`
 	Path        string   `json:"path" yaml:"path"`
 }
+
+// AppliesTo satisfies the ActionApplicable interface
+// by determining whether or not the Output applies to the provided action
+func (o *Output) AppliesTo(action string) bool {
```
silvin-lubecki: This method seems to be a duplicate of `func (p *Parameter) AppliesTo`, should we factorize it ?

vdice:
Different approach taken in https://github.com/deislabs/cnab-go/pull/129; closing.

PR #107, opened 2019-08-13
fix(bundle.go): add omitempty to contentDigest yaml tag

vdice:
Closes https://github.com/deislabs/cnab-go/issues/105


carolynvs:
The build passed but GH isn't getting the message. Merging with my magic powers.

PR #109, opened 2019-08-13
Properly handle optional credentials

### LINKED ISSUE #108: Duffle should not require a credential when it is optional
Consider the following duffle.json:

```
{
    "description": "A short description of your bundle",
    "credentials": {
        "username": {
            "description": "this is a cred",
            "env": "HOST_KEY"
        }
    },
	"invocationImages": {
		"cnab": {
			"builder": "docker",
			"configuration": {
				"registry": "jeremyrickard"
			},
			"name": "cnab"
		}
	},
	"keywords": ["test", "cnab", "tutorial"],
	"maintainers": [{
		"email": "john.doe@example.com",
		"name": "John Doe",
		"url": "https://example.com"
	}, {
		"email": "jane.doe@example.com",
		"name": "Jane Doe",
		"url": "https://example.com"
	}],
	"definitions": {
		"foo": {
			"type": "string"
		}
	},
	"parameters": {
		"foo": {
			"definition": "foo",
			"required": false,
			"destination": {
				"path": "",
				"env": "APP_URL"
			}
		},
		"manifest_file": {
            "definition": "foo",
            "required": false,
			"destination": {
                "path" : "",
				"env": "MANIFEST_FILE"
			}
		}
	},
	"name": "test",
	"schemaVersion": "v1.0.0-WD",
	"version": "0.1.0"
}
```

This produces a bundle that looks like:

```
{"credentials":{"username":{"description":"this is a cred","env":"HOST_KEY"}},"definitions":{"foo":{"type":"string"}},"description":"A short description of your bundle","invocationImages":[{"image":"jeremyrickard/test-cnab:7612a24756ad13e91aa1ff2d7795039045b43462","imageType":"docker"}],"keywords":["test","cnab","tutorial"],"maintainers":[{"email":"john.doe@example.com","name":"John Doe","url":"https://example.com"},{"email":"jane.doe@example.com","name":"Jane Doe","url":"https://example.com"}],"name":"test","parameters":{"foo":{"definition":"foo","destination":{"env":"APP_URL"}},"manifest_file":{"definition":"foo","destination":{"env":"MANIFEST_FILE"}}},"schemaVersion":"v1.0.0-WD","version":"0.1.0"}
```

In this case, the spec is clear that the credential _is not required_. We currently fail installation if _any_ credentials are defined though:

```
$ duffle install blah test:0.1.0
Error: bundle requires credential for username
```

Per the spec, we MUST fail if the credential is required. I believe if a credential is _optional_ we should fail fast. The bundle may or may not need the credential and we should defer to the bundle author to 
a.) tell the bundle consumers that it is _required_ by setting it to be required
b.) handle optional credentials in their bundle. 

This allows use cases like deislabs/duffle#574

jeremyrickard:
A credential has a `required` flag, which defaults to `true`. As #108 and the spec point out:

```
required indicates whether this credential MUST be supplied. By default it is false, which means the credential is optional. When true, a runtime MUST fail if the credential is not provided.
```

However, we were ALWAYS requiring any credential to have a value provided. This check occurred both in the CredentialSet Validate(..) method for that is used for stateful actions, but also the Expand(...) method. The Expand(..) method had an argument to flag an call to the method as stateless, in which case we allowed a credential to not be present but we still failed on optional for stateful calls. 

This PR adds explicit handling for optional credentials. 

Fixes #108

### DIFF
--- a/credentials/credentialset.go
+++ b/credentials/credentialset.go
@@ -25,7 +25,7 @@ func (s Set) Expand(b *bundle.Bundle, stateless bool) (env, files map[string]str
 	for name, val := range b.Credentials {
 		src, ok := s[name]
 		if !ok {
-			if stateless {
+			if stateless || !val.Required {
 				continue
 			}
 			err = fmt.Errorf("credential %q is missing from the user-supplied credentials", name)
@@ -77,15 +77,15 @@ func Load(path string) (*CredentialSet, error) {
 
 // Validate compares the given credentials with the spec.
 //
-// This will result in an error only if:
-// - a parameter in the spec is not present in the given set
-// - a parameter in the given set does not match the format required by the spec
+// This will result in an error only when the following conditions are true:
+// - a credential in the spec is not present in the given set
+// - the credential is required
 //
 // It is allowed for spec to specify both an env var and a file. In such case, if
-// the givn set provides either, it will be considered valid.
+// the givesn set provides either, it will be considered valid.
 func Validate(given Set, spec map[string]bundle.Credential) error {
-	for name := range spec {
-		if !isValidCred(given, name) {
+	for name, cred := range spec {
+		if !isValidCred(given, name) && cred.Required {
 			return fmt.Errorf("bundle requires credential for %s", name)
 		}
 	}
--- a/credentials/credentialset_test.go
+++ b/credentials/credentialset_test.go
@@ -117,7 +117,26 @@ func TestCredentialSet_Merge(t *testing.T) {
 
 }
 
-func TestCredentialSetMissingCred(t *testing.T) {
+func TestCredentialSetMissingRequiredCred(t *testing.T) {
+	b := &bundle.Bundle{
+		Name: "knapsack",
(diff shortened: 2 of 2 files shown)


glyn:
> A credential has a `required` flag, which defaults to `true`

Did you mean `false`?

credentials/credentialset.go
```diff
@@ -77,15 +77,15 @@ func Load(path string) (*CredentialSet, error) {
...
-// This will result in an error only if:
-// - a parameter in the spec is not present in the given set
-// - a parameter in the given set does not match the format required by the spec
+// This will result in an error only when the following conditions are true:
+// - a credential in the spec is not present in the given set
+// - the credential is required
 //
 // It is allowed for spec to specify both an env var and a file. In such case, if
-// the givn set provides either, it will be considered valid.
+// the givesn set provides either, it will be considered valid.
```
glyn: Nice try. How about this:
```suggestion
// the given set provides either, it will be considered valid.
```

PR #112, opened 2019-08-19
Support tweaking the operation before running it

carolynvs:
This will allow a tool to apply additional configuration, such as injecting additional files or environment variables, before the operation is run.

In our case, it will allow Porter to continue to implement dependencies without having to make extensive changes to cnab-go because it can inject the changes it requires through this configuration function instead of modifying the cnab-go code directly.


action/install.go
```diff
@@ -11,6 +11,10 @@ import (
 // Install describes an installation action
 type Install struct {
 	Driver driver.Driver // Needs to be more than a string
+
+	// OperationConfig is an optional handler that applies additional
+	// configuration to the operation before it is executed.
+	OperationConfig func(operation *driver.Operation)
```
glyn: I suggest defining a type for these functions to keep them all in step:
```
type OperationConfigFunc func(*driver.Operation) error
```
Note that it's probably worth allowing the handler to fail. For example, it might need to apply additional configuration which somehow contradicts what's already there because some constraint has been unexpectedly broken. Returning an error seems a better approach than requiring the function to panic and presenting a stack trace to the user.

glyn: Did you consider abstracting out a single struct and then have specific action structs extend it? For example:
```
// ConfigurableAction is used to define actions which can configure operations before they are executed.
type ConfigurableAction struct {
    // OperationConfig is an optional handler that applies additional
    // configuration to an operation before it is executed.        
    OperationConfig OperationConfigFunc
}
```

carolynvs: That's a good idea, I'll make that change. 👍

carolynvs: 👍

action/install.go
```diff
@@ -25,6 +29,10 @@ func (i *Install) Run(c *claim.Claim, creds credentials.Set, w io.Writer) error
 		return err
 	}
 
+	if i.OperationConfig != nil {
```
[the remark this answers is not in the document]
carolynvs: I originally made the change that way. 😀 It would be a breaking change since it alters the method signature. I wasn't sure if that would be a welcome change(?) so I went with this instead for an initial proposal.

It looked like this:

```
func (i *Install) Run(c *claim.Claim, creds credentials.Set, ops OperationConfig...)
```

Setting the writer on the op was something that the caller managed via the `ops` parameter.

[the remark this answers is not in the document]
silvin-lubecki: Ok I was wondering about that too, but as we didn't hit the 1.0 yet, should we "care" about about breaking changes? I think we should focus on the API design instead 😄

carolynvs:
I just dug up what duffle is doing (below) immediately before run and I think this would work for duffle as well. 

https://github.com/deislabs/duffle/blob/0c0cf660e05c549599056206214d9ead72b13bd7/cmd/duffle/main.go#L146-L156

Seems to me that duffle could add an config action with this function's logic and not need the wrapper.

carolynvs:
@glyn Ok, just pushed up another commit with your suggested changes.

action/install.go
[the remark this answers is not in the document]
carolynvs: Here's what it would look like with @silvin-lubecki 's suggested change. 

```
type OperationConfigFunc func(op *driver.Operation) error

type OperationConfigs []OperationConfigFunc

// ApplyConfig safely applies the configuration function to the operation, if defined,
// and returns any error.
func (cfgs OperationConfigs) ApplyConfig(op *driver.Operation) error {
	var err error
	for _, cfg := range cfgs {
		err = cfg(op)
		if err != nil {
			return err
		}
	}
	return nil
}

// Run performs an installation and updates the Claim accordingly
func (i *Install) Run(c *claim.Claim, creds credentials.Set, w io.Writer, opCfgs ...OperationConfigFunc) error {
	// ...

	err = OperationConfigs(opCfgs).ApplyConfig(op)
	if err != nil {
		return err
	}
```

I'm leaving the writer alone for now just to show what the overall change looks like. It could made into a config func that everyone would have to pass in (more disruptive), or we could keep it as is.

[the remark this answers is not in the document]
carolynvs: @glyn What do you think?

action/configurable_action_test.go
```diff
@@ -0,0 +1,46 @@
+package action
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/deislabs/cnab-go/driver"
```
radu-matei: Linter might not be happy.

action/configurable_action.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/deislabs/cnab-go/driver"
+)
+
+type OperationConfigFunc func(op *driver.Operation) error
+
+// ConfigurableAction is used to define actions which can configure operations before they are executed.
+type ConfigurableAction struct {
+	// OperationConfig is an optional handler that applies additional
+	// configuration to an operation before it is executed.
+	OperationConfig OperationConfigFunc
```
silvin-lubecki: I would have prefer a slice here, so user can have multiple unitary config modifiers. I guess it would help factorizing between different actions. WDYT?

action/run_custom_test.go
```diff
@@ -39,6 +39,44 @@ func TestRunCustom(t *testing.T) {
...
+		d := &mockDriver{
+			shouldHandle: true,
+			Result: driver.OperationResult{
+				Outputs: map[string]string{
+					"/tmp/some/path": "SOME CONTENT",
+				},
+			},
+			Error: nil,
+		}
+		inst := &RunCustom{Driver: d, Action: "test"}
```
silvin-lubecki: I think we should factorize those tests, as they are exactly the same for all the actions (good hint is that this action is still called `inst` 😇 ).

carolynvs:
@silvin-lubecki @glyn I've updated the PR with Silvin's requested change. If you could take another look, I'm hoping to get some form of this PR (either this or the former) merged before the cutoff. Thanks!

bundle/bundle_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"gopkg.in/yaml.v2"
+	yaml "gopkg.in/yaml.v2"
```
silvin-lubecki: nit: I think this alias is no-op, go already uses the part before the first dot as import name.

carolynvs: Yeah I'm aware it's a noop. It's an artifact of using GoLand for an IDE and for the life of me I can't figure out how to stop it from introducing those. When I remove the alias, it adds it back. 🤷‍♀  Running goimports (which I do on file save) doesn't clean those up either.

carolynvs: I just asked the twitterverse for help. https://twitter.com/carolynvs/status/1169993866503757827

PR #113, opened 2019-08-23
WIP: Introduce a digest validation interface and docker implementation

### LINKED ISSUE #101: Enforce digest validation
Currently, bundles don't have digests for invocation images (and other images). The deislabs/duffle#355 PR is skipping validation when there is no digest present. Bundles _should_ have digests and we should instead enforce the validation. This would be a breaking change at the moment (given state of bundles) so once deislabs/duffle#391 is completed, swap digest validation to be required, unless explicitly disabled.

jeremyrickard:
This introduces a Validate interface for image contentDigest validation, along with an implementation for Docker and OCI images that uses github.com/pivotal/image-relocation to obtain the digest based on the specified image image (thanks for the suggestion @glyn). This also allows us to implement additional validators down the road.

By default, validation will fail if a digest is not provided in the Image struct. This can be overridden by setting allowMissingDigests to true.

Note: this doesn't actually change any of the drivers/operation code. Clients of cnab-go will need to invoke the validation before calling an operation. This is done in order to allow validation of ALL the images referenced in the bundle. Currently, `Operation` only knows about the invocation image. This also allows clients to handle things like relocation mapping or any other mapping of bundle image to actual image that might be run. It also didn't feel appropriate to place the validation calls in the driver's themselves, as a given operation might actually need to validate multiple image types down the road (example: we use a Docker based invocation image that deploys OVA files to some VMware infrastructure) and it seemed like a cleaner separation to not mix with the drivers. Also allows this to be called outside of the context of a call to a driver Run(..), but I'm happy to wire it up to the Operations if desired. 

Closes #101

### DIFF
--- a/digests/validation.go
+++ b/digests/validation.go
@@ -1,0 +1,45 @@
+package digests
+
+import (
+	"fmt"
+
+	"github.com/deislabs/cnab-go/bundle"
+	"github.com/deislabs/cnab-go/digests/validators"
+	"github.com/pkg/errors"
+)
+
+// Validator validates that the contentDigest for each provided image matches
+// the authoritative source. By default, validation will fail if a contentDigest
+// is not provided for the image. Set allowMissingDigests to true to allow this
+type Validator interface {
+	Validate(images []bundle.Image, allowMissingDigests bool) error
+}
+
+type validator struct {
+	validators map[string]validators.ImageValidator
+}
+
+// NewValidator returns a digest Validator
+func NewValidator() Validator {
+	return &validator{
+		validators: validators.Create(),
+	}
+}
+
+func (v *validator) Validate(images []bundle.Image, allowMissingDigests bool) error {
+	for _, img := range images {
+		validator, ok := v.validators[img.ImageType]
+		if !ok {
+			return fmt.Errorf("unknown image type: %s", img.ImageType)
+		}
+		if img.Digest == "" && !allowMissingDigests {
+			return fmt.Errorf("unable to validate %s, digest not present", img.Image)
+		}
+		if img.Digest != "" {
+			if err := validator.Validate(img); err != nil {
+				return errors.Wrapf(err, "validation failed for %s", img.Image)
+			}
+		}
+	}
+	return nil
+}
--- a/digests/validation_test.go
+++ b/digests/validation_test.go
@@ -1,0 +1,147 @@
+package digests
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/deislabs/cnab-go/bundle"
+	"github.com/deislabs/cnab-go/digests/validators"
+	"github.com/stretchr/testify/assert"
+)
+
+type fakeImageValidator struct {
+	digest string
+}
+
+func (f fakeImageValidator) Validate(img bundle.Image) error {
+	if f.digest != img.Digest {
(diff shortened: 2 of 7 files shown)


digests/validators/docker_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (f fakeClient) NewLayout(path string) (registry.Layout, error) {
+	return f.layoutFake(path)
+}
+
+func (f fakeClient) ReadLayout(path string) (registry.Layout, error) {
+	return f.readLayoutFake(path)
+}
+
+func TestDigestMatches(t *testing.T) {
```
glyn: Suggestion: this could be combined with the next test function to form a table test.

digests/validators/validator.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/deislabs/cnab-go/bundle"
+)
+
+// Known image types for validation
+const (
+	ImageTypeDocker = "docker"
+	ImageTypeOCI    = "oci"
+)
+
+// Create builds a new image validator collection for know image types
```
glyn:
```suggestion
// Create builds a new image validator collection for known image types
```

digests/validators/validator.go
```diff
@@ -0,0 +1,26 @@
...
+func Create() map[string]ImageValidator {
+	d := NewDockerValidator()
+	return map[string]ImageValidator{
+		ImageTypeDocker: d,
+		ImageTypeOCI:    d,
+	}
+}
+
+// ImageValidator valdiates that the content digest for a given image matches
```
glyn:
```suggestion
// ImageValidator validates that the content digest for a given image matches
```

digests/validation.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func (v *validator) Validate(images []bundle.Image, allowMissingDigests bool) error {
+	for _, img := range images {
+		validator, ok := v.validators[img.ImageType]
+		if !ok {
+			return fmt.Errorf("unknown image type: %s", img.ImageType)
+		}
+		if img.Digest == "" && !allowMissingDigests {
```
glyn: I wonder if this `if` block should be moved before checking the existence of a validator? Then if `allowMissingDigests` is `true` and the digest is empty and the image type is unknown, validation would pass. Maybe it's better to be strict, but I could imagine that seeming pretty arbitrary if `allowMissingDigests` is `true`...

digests/validation_test.go
```diff
@@ -0,0 +1,147 @@
...
+		fmt.Sprintf(
+			"validation failed for %s: digest validation failed: expected %s found %s",
+			imageName,
+			expected,
+			actual,
+		),
+	)
+}
+
+func TestUnknownImageType(t *testing.T) {
```
glyn: Probably need to test unknown image type with and without a digest to clarify the intended behaviour. What that behaviour should be is another matter - see earlier comment. ;-)

digests/validation_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+func emptyValidator() fakeImageValidator {
+	return fakeImageValidator{
+		digest: "",
+	}
+}
+
+func TestNoImages(t *testing.T) {
```
glyn: This and the following tests look suitable to become a table test.

technosophos:
What should we do with this PR?

PR #114, opened 2019-08-26
Support pulling invocation images by digest

jlegrone:
This PR provides the information to driver implementations necessary to [reference invocation images by digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier), and implements pulling by digest for Kubernetes. Referencing by digest instead of tag is an inexpensive way to allow some drivers to ensure that the correct invocation image digest is resolved.

Note that this does _not_ provide any validation of image digests beyond whatever is supported by the user's container runtime, and is complementary to the work in https://github.com/deislabs/cnab-go/pull/113.

It looks like Kubernetes [supports digest validation](https://github.com/kubernetes/kubernetes/blob/d999f456ea51e4ce39af599e1ade66f8da75fd91/pkg/kubelet/dockershim/libdocker/helpers.go#L99-L125), so offloading this to the container runtime is a possible solution to #101 in some cases.


testdata/operations/valid-operation.json
```diff
@@ -1,18 +1,20 @@
...
-    },
+  "installation_name": "test",
+  "revision": "01DDY0MT808KX0GGZ6SMXN4TW",
+  "action": "install",
+  "parameters": {
+    "param1": "value1",
+    "param2": "value2"
+  },
+  "image": {
     "image": "testing.azurecr.io/duffle/test:e8966c3c153a525775cbcddd46f778bed25650b4",
```
jlegrone: This image needs auth to pull, otherwise I'd like to add a digest to the test file.

jlegrone: Updated to use a public sample invocation image in https://github.com/deislabs/cnab-go/pull/114/commits/3ad2f2446805c4b3c02601a3df52ad284136f8fb

PR #118, opened 2019-09-04
fix(bundle/definition): add custom validator for contentEncoding

vdice:
This represents a fix to ensure definitions with a non-empty `contentEncoding` property pass validation by adding a custom validator for continued use of the jsonschema library cnab-go currently uses.  It's specific to `contentEncoding` but illustrates a potential path forward for further "custom" validators (either for properties in the default JSON Schema but *not* stock in the jsonschema library or for truly custom properties/validators down the line.)

See https://github.com/deislabs/cnab-go/issues/117 for background on general issue and https://github.com/deislabs/porter/pull/579 for feature that uses the changes here to pass validation.

### DIFF
--- a/bundle/definition/validation_test.go
+++ b/bundle/definition/validation_test.go
@@ -42,6 +42,39 @@ func TestObjectValidationValid(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestObjectValidationValid_CustomValidator(t *testing.T) {
+	s := `{
+		"type": "object",
+		"properties" : {
+			"file" : {
+				"type": "string",
+				"contentEncoding": "base64"
+			}
+		},
+		"required" : ["file"]
+	}`
+	definition := new(Schema)
+	err := json.Unmarshal([]byte(s), definition)
+	require.NoError(t, err, "should have been able to marshal definition")
+	assert.Equal(t, "object", definition.Type, "type should have been an object")
+	props := definition.Properties
+	assert.NotNil(t, props, "should have found properties")
+	assert.Equal(t, 1, len(props), "should have had a single property")
+	propSchema, ok := props["file"]
+	assert.True(t, ok, "should have found file property")
+	assert.Equal(t, "string", propSchema.Type, "file type should have been a string")
+	assert.Equal(t, "base64", propSchema.ContentEncoding, "file contentEncoding should have been base64")
+
+	val := struct {
+		File string `json:"file"`
+	}{
+		File: "SGVsbG8gV29ybGQhCg==",
+	}
+	valErrors, err := definition.Validate(val)
+	assert.Len(t, valErrors, 0, "expected no validation errors")
+	assert.NoError(t, err)
+}
+
 func TestObjectValidationInValidMinimum(t *testing.T) {
 	s := `{
 		"type": "object",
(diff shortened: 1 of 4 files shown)


bundle/definition/validators.go
```diff
@@ -0,0 +1,29 @@
...
+// Validate implements the Validator interface for ContentEncoding
+// Currently, this is a no-op and is only used to register with the jsonschema library
+// that 'contentEncoding' is a valid property (as of writing, it isn't one of the defaults)
+func (c ContentEncoding) Validate(propPath string, data interface{}, errs *[]jsonschema.ValError) {}
+
+// NewRootSchema returns a jsonschema.RootSchema with any needed custom
+// jsonschema.Validators pre-registered
+func NewRootSchema() *jsonschema.RootSchema {
+	// Register custom validators here
+	// Note: as of writing, jsonschema doesn't have a stock validator for intances of type `contentEncoding`
```
glyn:
```suggestion
	// Note: as of writing, jsonschema doesn't have a stock validator for instances of type `contentEncoding`
```

bundle/definition/validation_test.go
```diff
@@ -42,6 +42,39 @@ func TestObjectValidationValid(t *testing.T) {
...
+	assert.Equal(t, 1, len(props), "should have had a single property")
+	propSchema, ok := props["file"]
+	assert.True(t, ok, "should have found file property")
+	assert.Equal(t, "string", propSchema.Type, "file type should have been a string")
+	assert.Equal(t, "base64", propSchema.ContentEncoding, "file contentEncoding should have been base64")
+
+	val := struct {
+		File string `json:"file"`
+	}{
+		File: "SGVsbG8gV29ybGQhCg==",
```
glyn: If I change this to an invalid base64 value such as "SGVsbG8gV29ybGQhCg===", the test still passes. Shouldn't it fail?

bundle/definition/validators.go
```diff
@@ -0,0 +1,29 @@
+package definition
+
+import (
+	"github.com/qri-io/jsonschema"
+)
+
+// ContentEncoding represents a "custom" Schema property
+type ContentEncoding string
```
silvin-lubecki: As `NewContentEncoding` is returning a `jsonschema.Validator`, do we really need to expose that type?

bundle/definition/validation_test.go
vdice: Good call; indeed, as mentioned in the code comments, the current validator is a no-op (simply _allows_ for a `contentEncoding` property), but we can and perhaps should actually perform some basic validations specific to this property.  Preference on doing so in this PR or as a follow-up?

glyn: Actually, I'd prefer it in this PR to check feasibility _before_ merging this PR. Hope it's not too much work - maybe a separate commit?

glyn: I think the values we need to support are defined by https://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3.

vdice: Ok, validation added for `base64`.  Wanted to keep scope limited so didn't venture off into adding other content encoding validations quite yet.  What do you think?

glyn: Fair enough. Why not raise issue(s) for the other encodings?

vdice: Deal! https://github.com/deislabs/cnab-go/issues/132

bundle/definition/validators.go
silvin-lubecki: I don't have strong opinion on it, I just have the habit of reducing the exposed parts to their minimum 👍

PR #119, opened 2019-09-04
Allow installation names including "." character

jlegrone:
From the [docs](https://github.com/deislabs/cnab-spec/blob/cc78ad5f2f7f4fabce6247f02cafb786bc011e07/103-bundle-runtime.md#environment-variables):

> Installation names MUST consist of Graph Unicode characters and MAY be user-readable. The Unicode Graphic characters include letters, marks, numbers, punctuation, symbols, and spaces, from categories L, M, N, P, S, Zs.

This PR doesn't quite bring cnab-go up to spec, but does enable the use of installation names which include the `.` character (this supports our internal installation naming convention). If someone would like to school me in regex we can update this PR to be fully spec compliant. 😄

### DIFF
--- a/claim/claim_test.go
+++ b/claim/claim_test.go
@@ -45,18 +45,23 @@ func TestUpdate(t *testing.T) {
 }
 
 func TestValidName(t *testing.T) {
-	is := assert.New(t)
-
 	for name, expect := range map[string]bool{
 		"M4cb3th":               true,
-		"Lady MacBeth":          false, //spaces illegal
+		"Lady MacBeth":          false, // spaces illegal
 		"3_Witches":             true,
 		"Banquø":                false, // We could probably loosen this one up
 		"King-Duncan":           true,
 		"MacDuff@geocities.com": false,
 		"hecate":                true, // I wouldn't dare cross Hecate.
+		"foo bar baz":           false,
+		"foo.bar.baz":           true,
+		"foo-bar-baz":           true,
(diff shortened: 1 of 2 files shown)


claim/claim_test.go
```diff
@@ -45,18 +45,23 @@ func TestUpdate(t *testing.T) {
...
 
 func TestValidName(t *testing.T) {
-	is := assert.New(t)
-
 	for name, expect := range map[string]bool{
 		"M4cb3th":               true,
-		"Lady MacBeth":          false, //spaces illegal
+		"Lady MacBeth":          false, // spaces illegal
 		"3_Witches":             true,
 		"Banquø":                false, // We could probably loosen this one up
```
SantiagoTorres: Unrelated to this PR, but `ø` is part of L and here it isn't approved. Group S is spacing and is also approved. In fact, the only group that seems to not be allowed here is `C` so none of the tests here should fail. Am I missing something?

SantiagoTorres: On that note, my understanding is that, if you are using `regex` (and thus `re2`) you can get away using unicode character classes explicitly and probably get away with `[\pL\pM\pN\pP\pS\pZ]*` or so...

jlegrone: I'll open a new issue for this, thanks!

jlegrone: See #126

PR #124, opened 2019-09-16
Add Urvashi Reddy as a cnab-go maintainer

jeremyrickard:
Pivotal's contributions for outputs in cnab-go and the Duffle driver were very important and they  have indicated they will continue to be involved in the spec, as well as Duffle and cnab-go development, so I'd like to nominate Urvashi  as a maintainer.


jeremyrickard:
We have 5 LGTMs, also I say LGTM (so 6), so welcome to the maintainers @youreddy!

PR #129, opened 2019-09-19
feat(*): populate applicable output values with defaults when necessary

vdice:
If an applicable output (read: applies to a given action) is missing after the operation completes, supply an entry with the output's default value, if/when the default exists.

Note that this has only been added to the Docker driver which appears to be the sole driver implementing outputs as of writing.

Closes https://github.com/deislabs/cnab-go/issues/122

### DIFF
--- a/bundle/outputs.go
+++ b/bundle/outputs.go
@@ -6,3 +6,17 @@ type Output struct {
 	Description string   `json:"description,omitempty" yaml:"description,omitempty"`
 	Path        string   `json:"path" yaml:"path"`
 }
+
+// AppliesTo returns a boolean value specifying whether or not
+// the Output applies to the provided action
+func (output *Output) AppliesTo(action string) bool {
+	if len(output.ApplyTo) == 0 {
+		return true
+	}
+	for _, act := range output.ApplyTo {
+		if action == act {
+			return true
+		}
+	}
+	return false
+}
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	unix_path "path"
+	"strings"
 
 	"github.com/deislabs/cnab-go/driver"
 	"github.com/docker/cli/cli/command"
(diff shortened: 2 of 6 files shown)


driver/docker/docker.go
```diff
@@ -301,9 +302,35 @@ func (d *Driver) fetchOutputs(ctx context.Context, container string, op *driver.
 		return opResult, err
 	}
 
+	// if an applicable output is expected but does not exist and it has a
+	// non-empty default value, create an entry in the map with the
+	// default value as its contents
+	for name, output := range op.Bundle.Outputs {
```
carolynvs: One part that I'm unsure about is whether or not it is sufficient to just write to the output map? The spec says

> A runtime can leverage appropriate in-memory volume mounted at the path location for storing these outputs.

If we just write to the map, then the mounted volume will be missing any outputs that have a default, and only the claim (which this map is written) will have the full set of outputs. Having said that I'm not sure where the best place would be to write out the files with the defaults so that they both agree however.

Thoughts?

vdice: Hmmm, good point.  I'm also unclear on how to address this.

It feels like the outputs written -- or not written -- to the container's filesystem would be under the bundle author's domain and so my first thought is that it'd be odd for the library to write missing/additional files there.  Instead, adding/supplying default values to the operation/claim as a runtime responsibility seems more natural.  However, this is just my initial take and is moot if it differs from the spec.

vdice:
The most recent commit was to address the following bit in the spec: `Otherwise the runtime should report the missing output as an error to the user.` https://github.com/deislabs/cnab-go/pull/129/commits/187c8aa8948e2e966c91ed183815636b88e585ad

driver/docker/docker.go
```diff
@@ -301,9 +302,37 @@ func (d *Driver) fetchOutputs(ctx context.Context, container string, op *driver.
...
+	}
+
 	return opResult, nil
 }
 
+func existsInOutputsMap(outputsMap map[string]string, output string) bool {
+	var exists bool
+	for outputPath := range outputsMap {
+		if strings.Contains(outputPath, output) {
+			exists = true
```
silvin-lubecki: I guess you can early return here? `return true`, and at the end `return false` ?

driver/docker/docker.go
```diff
@@ -301,9 +302,37 @@ func (d *Driver) fetchOutputs(ctx context.Context, container string, op *driver.
...
+					return opResult, fmt.Errorf("required output %s is missing and has no default", name)
+				}
+			}
+		}
+	}
+
 	return opResult, nil
 }
 
+func existsInOutputsMap(outputsMap map[string]string, output string) bool {
```
silvin-lubecki: nit: rename `output` to `path` ?

driver/docker/docker.go
```diff
@@ -301,9 +302,37 @@ func (d *Driver) fetchOutputs(ctx context.Context, container string, op *driver.
...
+	var exists bool
+	for outputPath := range outputsMap {
+		if strings.Contains(outputPath, output) {
```
silvin-lubecki: I `string.Contains` enough? Or do we want to check the exact path?

bundle/outputs.go
```diff
@@ -6,3 +6,17 @@ type Output struct {
 	Description string   `json:"description,omitempty" yaml:"description,omitempty"`
 	Path        string   `json:"path" yaml:"path"`
 }
+
+// AppliesTo returns a boolean value specifying whether or not
+// the Output applies to the provided action
+func (output *Output) AppliesTo(action string) bool {
+	if len(output.ApplyTo) == 0 {
```
silvin-lubecki: It feels like there's a pattern with the Parameter type, I think we should factorize the code, maybe using embedded type?

driver/docker/docker.go
vdice: Good call; checking exact path now.

bundle/outputs.go
vdice: It's true... I could use some pointers/examples here.  I've tried a couple different ways (couldn't figure out how to prevent embedded struct from printing if empty on Marshal; couldn't quite figure out how to use an interface to avoid code duplication) but had to admit defeat.  Help would be welcome!  Could be in a follow-up as well, if that'd work.

silvin-lubecki: Ok let's keep it like this for now 👍

PR #130, opened 2019-09-19
Add support for outputs to command driver

### LINKED ISSUE #84: [spec compliance] Implement Outputs in Command Driver
The Command driver should implement output handling.

Since the command driver works by exec-ing a local binary, maybe the way to get output files would be to agree on a folder where the binary would leave them. I'm not sure where the interface between the Command driver and the binary it calls is defined.

simongdavies:
Adds support for outputs to the command driver.

Command driver will create a temp folder for the driver process to write outputs to and pass the folder path to the driver process in the  environment variable CNAB_OUTPUT_DIR

fixes #84

### DIFF
--- a/driver/command/command.go
+++ b/driver/command/command.go
@@ -130,26 +130,11 @@ func (d *Driver) getOperationResult(op *driver.Operation) (driver.OperationResul
 	opResult := driver.OperationResult{
 		Outputs: map[string]string{},
 	}
-	if len(op.Outputs) == 0 {
-		return opResult, nil
-	}
-
 	for _, item := range op.Outputs {
-		// Check if a value is required for this output and get the default value if available
-		valueRequired, defaultValue, err := checkIfOutputValueRequired(item, op, opResult)
-		if err != nil {
-			return opResult, fmt.Errorf("Command driver (%s) failed checking if output value required for item: %s Error: %v", d.Name, item, err)
-		}
-
 		fileName := path.Join(d.outputDirName, item)
-		_, err = os.Stat(fileName)
+		_, err := os.Stat(fileName)
 		if err != nil {
 			if os.IsNotExist(err) {
-				// Set a default value if available and required otherwise fail
-				if valueRequired {
-					return opResult, fmt.Errorf("Command driver (%s) failed for item: %s no output value found and no default value set", d.Name, item)
-				}
-				opResult.Outputs[item] = defaultValue
 				continue
 			}
 			return opResult, fmt.Errorf("Command driver (%s) failed checking for output file: %s Error: %v", d.Name, item, err)
@@ -162,13 +147,17 @@ func (d *Driver) getOperationResult(op *driver.Operation) (driver.OperationResul
 
 		opResult.Outputs[item] = string(contents)
 	}
+	// Check if there are missing outputs and get default values if any
+	for name, output := range op.Bundle.Outputs {
+		if output.AppliesTo(op.Action) {
+			if _, exists := opResult.Outputs[output.Path]; !exists {
+				if outputDefinition, exists := op.Bundle.Definitions[output.Definition]; exists && outputDefinition.Default != nil {
+					opResult.Outputs[output.Path] = fmt.Sprintf("%v", outputDefinition.Default)
+				} else {
+					return opResult, fmt.Errorf("Command driver (%s) failed - required output %s for action %s is missing and has no default is defined", d.Name, name, op.Action)
+				}
+			}
+		}
+	}
 	return opResult, nil
-}
-func checkIfOutputValueRequired(item string, op *driver.Operation, opResult driver.OperationResult) (bool, string, error) {
-	// TODO Check if the output is required by the action and write test to validate (requires update to op definition in https://github.com/deislabs/cnab-go/pull/129 to be merged )
-	// Until this check is inmplemented then all values are required for all bundles and actions
-	// Output default values and applies to are ignored by this driver until this function is implemented
-	// Function should return false if a output does not apply to an action or there is a default value (which it should return as the second return value)
-	return true, "", nil
-
 }
--- a/driver/command/command.go
+++ b/driver/command/command.go
@@ -5,8 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
 
 	"github.com/deislabs/cnab-go/driver"
@@ -14,7 +16,8 @@ import (
 
 // Driver relies upon a system command to provide a driver implementation
 type Driver struct {
-	Name string
+	Name          string
+	outputDirName string
 }
 
 // Run executes the command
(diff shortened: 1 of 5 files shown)


driver/command/command.go
```diff
@@ -101,5 +119,49 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+			return opResult, fmt.Errorf("Command driver (%s) failed checking for output file: %s Error: %v", d.Name, item, err)
+		}
+
+		file, err := os.Open(fileName)
+		if err != nil {
+			return opResult, fmt.Errorf("Command driver (%s) failed opening output file: %s Error: %v", d.Name, item, err)
+		}
+
+		defer file.Close()
+		contents, err := ioutil.ReadAll(file)
```
carolynvs: You can combine `os.Open`, `ioutil.ReadAll` and `file.Close` into a single call with `ioutil.ReadFile(fileName)`

driver/command/command.go
```diff
@@ -101,5 +119,49 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+		if err != nil {
+			return opResult, fmt.Errorf("Command driver (%s) failed writing output file: %s Error: %v", d.Name, item, err)
```
carolynvs: The error message says "error writing output" but it really failed on reading the output, no?

driver/command/command_nix_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/deislabs/cnab-go/driver"
+	"github.com/stretchr/testify/assert"
+
+	"os"
+	"testing"
+)
+
+func TestCommandDriverOutputs(t *testing.T) {
+	content := `#!/bin/sh
+		mkdir -p "${CNAB_OUTPUT_DIR}//cnab/app/outputs"
```
carolynvs: nit: I don't think you don't need the double `//` after CNAB_OUTPUT_DIR, just 1 slash.

driver/command/command.go
```diff
@@ -101,5 +119,49 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+
+	if len(op.Outputs) == 0 {
+		return opResult, nil
+	}
+
+	for _, item := range op.Outputs {
+		fileName := path.Join(d.outputDirName, item)
+		_, err := os.Stat(fileName)
+		if err != nil {
+			if os.IsNotExist(err) {
```
carolynvs: Can you add a test case that exercises this path? When an output file wasn't created?

simongdavies:
Addressed the comments, implemented a place holder to check for defaults once #129 is merged

driver/command/command_test.go
```diff
@@ -20,27 +20,36 @@ func TestCheckDriverExists(t *testing.T) {
...
 	name = "existing-driver"
-	cmddriver = &Driver{Name: name}
+	testfunc := func(t *testing.T, cmddriver *Driver) {
+		if !cmddriver.CheckDriverExists() {
+			t.Fatalf("Expected driver %s to exist", cmddriver.Name)
+		}
+	}
+	CreateAndRunTestCommandDriver(t, name, "", testfunc)
+}
+func CreateAndRunTestCommandDriver(t *testing.T, name string, content string, testfunc func(t *testing.T, d *Driver)) {
```
silvin-lubecki: nit: empty line

driver/command/command.go
```diff
@@ -101,5 +119,62 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+		if err != nil {
+			return opResult, fmt.Errorf("Command driver (%s) failed reading output file: %s Error: %v", d.Name, item, err)
+		}
+
+		opResult.Outputs[item] = string(contents)
+	}
+
+	return opResult, nil
+}
+func checkIfOutputValueRequired(item string, op *driver.Operation, opResult driver.OperationResult) (bool, string, error) {
```
silvin-lubecki: nit: empty line

driver/command/command.go
```diff
@@ -101,5 +119,62 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+	}
+
+	result, err := d.getOperationResult(op)
+	if err != nil {
+		return driver.OperationResult{}, fmt.Errorf("Command driver (%s) failed getting operation result: %v", d.Name, err)
+	}
+
+	return result, nil
+}
+func (d *Driver) getOperationResult(op *driver.Operation) (driver.OperationResult, error) {
```
silvin-lubecki: nit: empty line

driver/command/command.go
```diff
@@ -101,5 +119,62 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+	opResult := driver.OperationResult{
+		Outputs: map[string]string{},
+	}
+
+	if len(op.Outputs) == 0 {
```
silvin-lubecki: I guess this check is not useful as if `Outputs` is empty, ranging on it won't do anything?

driver/command/command.go
```diff
@@ -101,5 +119,62 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+
+	return opResult, nil
+}
+func checkIfOutputValueRequired(item string, op *driver.Operation, opResult driver.OperationResult) (bool, string, error) {
+	// TODO Check if the output is required by the action and write test to validate (requires update to op definition in https://github.com/deislabs/cnab-go/pull/129 to be merged )
+	// Until this check is inmplemented then all values are required for all bundles and actions
+	// Output default values and applies to are ignored by this driver until this function is implemented
+	// Function should return false if a output does not apply to an action or there is a default value (which it should return as the second return value)
+	return true, "", nil
+
```
silvin-lubecki: nit: empty line

driver/command/command.go
```diff
@@ -101,5 +119,62 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
...
+			return opResult, fmt.Errorf("Command driver (%s) failed reading output file: %s Error: %v", d.Name, item, err)
+		}
+
+		opResult.Outputs[item] = string(contents)
+	}
+
+	return opResult, nil
+}
+func checkIfOutputValueRequired(item string, op *driver.Operation, opResult driver.OperationResult) (bool, string, error) {
+	// TODO Check if the output is required by the action and write test to validate (requires update to op definition in https://github.com/deislabs/cnab-go/pull/129 to be merged )
```
silvin-lubecki: I wonder if we shouldn't mark this PR as `[WIP]` as it depends on another PR. I feel a bit uncomfortable merging the code as is, with this unimplemented feature.

simongdavies: Yeah buts its unimplemented everywhere until #129 is merged, once that is done its going to be straightforward to implement here, alternative I can remove this placeholder function

driver/command/command.go
simongdavies: Not sure I follow , function returns if outputs is empty?

silvin-lubecki: Yes but just after you iterate on `op.Outputs`, so if it is empty it will be no-op right? And then return at the end of the function?

simongdavies: I realized what you meant after I replied, I was just trying to be explicit about it, anyway that code has gone now it checks for default values.

driver/command/command.go
simongdavies: This is implemented now

driver/command/command_nix_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+package command
+
+import (
+	"github.com/deislabs/cnab-go/bundle"
+	"github.com/deislabs/cnab-go/bundle/definition"
+	"github.com/deislabs/cnab-go/driver"
+	"github.com/stretchr/testify/assert"
+
+	"os"
```
silvin-lubecki: Nit: this import layout looks weird, the standard library imports should be first

PR #135, opened 2019-09-26
Refactor export logic out of Duffle

jeremyrickard:
This PR moves code from Duffle to cnab-go related to export operations, but doesn't make an effort to refactor/rewrite any of it.


Gopkg.lock
```diff
@@ -199,6 +199,17 @@
   revision = "8991bc29aa16c548c550c7ff78260e27b9ab7c73"
   version = "v1.1.1"
 
+[[projects]]
+  digest = "1:5322d845714997eb15bc0f521bd40a53988337606798407ec99281b1554ead22"
+  name = "github.com/deislabs/duffle"
```
silvin-lubecki: I think we are introducing a cyclic dependency here 🤔

Gopkg.lock
```diff
@@ -1156,19 +1204,25 @@
   input-imports = [
     "github.com/Azure/go-autorest/autorest",
     "github.com/Masterminds/semver",
+    "github.com/deislabs/duffle/pkg/imagestore",
```
glyn: I think this package needs copying too to break the cyclic dependency.

packager/export.go
```diff
@@ -0,0 +1,173 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/deislabs/cnab-go/bundle"
+	"github.com/deislabs/cnab-go/bundle/loader"
+	"github.com/docker/docker/pkg/archive"
+
+	"github.com/deislabs/duffle/pkg/imagestore"
```
radu-matei: This should be imported from `deislabs/cnab-go`.

packager/export_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"path/filepath"
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/deislabs/cnab-go/bundle/loader"
+
+	"github.com/deislabs/duffle/pkg/imagestore"
+	"github.com/deislabs/duffle/pkg/imagestore/imagestoremocks"
```
radu-matei: This should be imported from `deislabs/cnab-go`.

glyn: We need to decide whether to copy across the two image store implementations too or whether to leave them and the constructor functions in duffle. I don't feel strongly - it might be simplest to copy the lot.

packager/export.go
jeremyrickard: Argh, I thought I had updated that! Sorry!, will amend,

packager/export_test.go
jeremyrickard: I think they need to be copied over.

jeremyrickard:
OOPS! Sorry I thought I had updated all the duffle refs. Must have screwed a stash or something. fixing...

packager/export_test.go
radu-matei: This generates a cyclical import if left like this.

jeremyrickard:
> We need to decide whether to copy across the two image store implementations too or whether to leave them and the constructor functions in duffle. I don't feel strongly - it might be simplest to copy the lot.

@glyn raised this in a comment above and i didn't want it to get lost. I think they should come over, the oci layout at least should come across. It's likely that both would be useful. relocation depends on the remote store if i recall, and i think the relocation functionality might be something that is reusable (maybe we don't put it in cnab-go but make a cnab-relocation library or something that can be shared?)

glyn:
> > We need to decide whether to copy across the two image store implementations too or whether to leave them and the constructor functions in duffle. I don't feel strongly - it might be simplest to copy the lot.
> 
> @glyn raised this in a comment above and i didn't want it to get lost. I think they should come over, the oci layout at least should come across. It's likely that both would be useful. relocation depends on the remote store if i recall, and i think the relocation functionality might be something that is reusable (maybe we don't put it in cnab-go but make a cnab-relocation library or something that can be shared?)

I think it's ok in cnab-go for now.

packager/import.go
```diff
@@ -0,0 +1,91 @@
...
+		Loader:      load,
+		Verbose:     verbose,
+	}, nil
+}
+
+// Import decompresses a bundle from Source (location of the compressed bundle) and properly places artifacts in the correct location(s)
+func (im *Importer) Import() error {
+	_, _, err := im.Unzip()
+
+	// TODO: https://github.com/deislabs/duffle/issues/758
```
silvin-lubecki: Should we add an issue in cnab-go instead to track it?

jeremyrickard:
Looks like brigade isn't running tests for the PR? @vdice

packager/import.go
jeremyrickard: should i just move that duffle issue over here?

vdice:
@jeremyrickard there indeed was a DNS hiccup yesterday (so GH webhooks weren't getting to our Brigade gateway).  Please push an amended commit when convenient; it'll re-trigger.

packager/import.go
glyn: Moved.

PR #137, opened 2019-10-01
Fix for missing k8s logs

jzh0u:
Fix for the missing pod logs in the kubernetes driver.

We didn't check to make sure io.Copy() was writing anything, so sometimes it would finish before the pod actually wrote anything. Added a few checks to remedy this.

To reproduce bug:
Run `go test -tags=integration -count=20 ./driver/kubernetes`
See below fail at least once
image

### DIFF
--- a/driver/kubernetes/kubernetes.go
+++ b/driver/kubernetes/kubernetes.go
@@ -3,11 +3,10 @@ package kubernetes
 import (
 	"fmt"
 	"io"
-	"math/rand"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
+	"time"
 
 	// load credential helpers
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
@@ -30,6 +29,7 @@ import (
 const (
 	k8sContainerName    = "invocation"
 	k8sFileSecretVolume = "files"
+	numBackoffLoops     = 6
 )
 
 // Driver runs an invocation image in a Kubernetes cluster.
@@ -232,22 +232,29 @@ func (k *Driver) Run(op *driver.Operation) (driver.OperationResult, error) {
 		return driver.OperationResult{}, nil
 	}
 
-	selector := metav1.ListOptions{
+	// Create a selector to detect the job just created
+	jobSelector := metav1.ListOptions{
 		LabelSelector: labels.Set(job.ObjectMeta.Labels).String(),
+		FieldSelector: newSingleFieldSelector("metadata.name", job.ObjectMeta.Name),
 	}
 
-	return driver.OperationResult{}, k.watchJobStatusAndLogs(selector, op.Out)
+	// Prevent detecting pods from prior jobs by adding the job name to the labels
+	podSelector := metav1.ListOptions{
+		LabelSelector: newSingleFieldSelector("job-name", job.ObjectMeta.Name),
+	}
+
+	return driver.OperationResult{}, k.watchJobStatusAndLogs(podSelector, jobSelector, op.Out)
 }
 
-func (k *Driver) watchJobStatusAndLogs(selector metav1.ListOptions, out io.Writer) error {
+func (k *Driver) watchJobStatusAndLogs(podSelector metav1.ListOptions, jobSelector metav1.ListOptions, out io.Writer) error {
 	// Stream Pod logs in the background
 	logsStreamingComplete := make(chan bool)
-	err := k.streamPodLogs(selector, out, logsStreamingComplete)
+	err := k.streamPodLogs(podSelector, out, logsStreamingComplete)
 	if err != nil {
 		return err
 	}
 	// Watch job events and exit on failure/success
-	watch, err := k.jobs.Watch(selector)
+	watch, err := k.jobs.Watch(jobSelector)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


driver/kubernetes/kubernetes.go
```diff
@@ -316,8 +316,15 @@ func (k *Driver) streamPodLogs(options metav1.ListOptions, out io.Writer, done c
...
-			io.Copy(out, reader)
+			n, err := io.Copy(out, reader)
 			reader.Close()
+			if err != nil {
+				continue
+			}
+			if n == 0 {
+				streamedLogs[podName] = false
+				continue
+			}
```
jlegrone: This may cause an infinite loop if an invocation image does not have any output. Can we retry up to a certain number of times to get logs, and then exit?

driver/kubernetes/kubernetes.go
```diff
@@ -316,8 +316,15 @@ func (k *Driver) streamPodLogs(options metav1.ListOptions, out io.Writer, done c
 			// We successfully connected to the pod, so mark it as having streamed logs.
 			streamedLogs[podName] = true
 			// Block the loop until all logs from the pod have been processed.
-			io.Copy(out, reader)
+			n, err := io.Copy(out, reader)
 			reader.Close()
+			if err != nil {
+				continue
```
glyn: If there's a (transient) error, it is possible that no logs were copied and yet we continue with the pod marked as having streamed logs, so a retry won't occur.

driver/kubernetes/kubernetes.go
```diff
@@ -316,8 +316,15 @@ func (k *Driver) streamPodLogs(options metav1.ListOptions, out io.Writer, done c
 			// We successfully connected to the pod, so mark it as having streamed logs.
 			streamedLogs[podName] = true
```
glyn: Wouldn't it be better to set this to true if and only if (all?) the logs have been streamed?

driver/kubernetes/kubernetes.go
```diff
@@ -346,6 +368,7 @@ func generateLabels(op *driver.Operation) map[string]string {
 		"cnab.io/installation": op.Installation,
 		"cnab.io/action":       op.Action,
 		"cnab.io/revision":     op.Revision,
+		"random_string":        strconv.Itoa(rand.Intn(1000)),
```
glyn: How is this relevant to this PR?

driver/kubernetes/kubernetes.go
```diff
@@ -291,34 +293,54 @@ func (k *Driver) streamPodLogs(options metav1.ListOptions, out io.Writer, done c
...
-				Follow:    true,
-			})
-			reader, err := req.Stream()
-			// There was an error connecting to the pod, so continue the loop and attempt streaming
-			// logs again next time there is an event for the same pod.
-			if err != nil {
-				continue
+
+			timeout := 0
+			for (streamedLogs[podName] < numBackoffLoops) && (timeout < numBackoffLoops) {
```
glyn: Unfortunately, the additional logic is becoming more convoluted rather than simpler.

driver/kubernetes/kubernetes.go
```diff
@@ -291,34 +293,54 @@ func (k *Driver) streamPodLogs(options metav1.ListOptions, out io.Writer, done c
...
+				req := k.pods.GetLogs(podName, &v1.PodLogOptions{
+					Container: k8sContainerName,
+					Follow:    true,
+				})
+				reader, err := req.Stream()
+				// There was an error connecting to the pod, so continue the loop and attempt streaming
+				// logs again next time there is an event for the same pod.
+				if err != nil {
+					timeout = timeout + 1
```
glyn: There isn't a delay before trying again, so a transient problem might not have chance to clear.

driver/kubernetes/kubernetes.go
```diff
@@ -291,34 +293,54 @@ func (k *Driver) streamPodLogs(options metav1.ListOptions, out io.Writer, done c
...
+				// Block the loop until all logs from the pod have been processed.
+				n, err := io.Copy(out, reader)
+				reader.Close()
+				if err != nil {
+					continue
+				}
+				// There is a chance where we have connected to the pod, but it has yet to write something.
+				// In that case, we continue to to keep streaming until it does.
+				if n == 0 {
+					continue
```
glyn: Here we _know_ we haven't copied any logs, but `streamedLogs[podName]` has still been incremented, which seems incorrect.

glyn:
@jzh0u Thanks for trying again, but the logic is becoming pretty convoluted. It might be best to try to strip it down to the absolute minimum required to fix the problem.

Also, regardless of this PR, you've found a valid bug, so it would probably be helpful to submit an issue.

driver/kubernetes/kubernetes.go
jzh0u: It's a temporary fix for a pod race condition I was running into. Sometimes, I ended up streaming previous pods that haven't finished terminating because the labels are the same between trials. I'm currently looking into a better way of doing this, but just wanted to quickly fix the issue.

driver/kubernetes/kubernetes.go
jzh0u: But in the case that there are no logs, if we don't increment `streamedLogs[podName]` the whole thing will hang

glyn:
@vdice Please could you advise what's gone wrong with the `tests` check? I can't find any way to debug this.

PR #139, opened 2019-10-02
Add Docker driver config to keep container around

### LINKED ISSUE #128: Docker driver should have parameter to keep invocation containers
When a bundle developer is working on their bundle, they will try it out with `duffle install` and probably use the docker driver to run their invocation image locally. After a failed run (or sometimes after a seemingly successful one), developers are likely to want to poke around at the state of the invocation image at the end of the run. Since the CNAB runtime injects a number of files into the image, they want to explore the actual state of the container that ran.

Currently, the docker driver always cleans up the containers, so this is not possible. A new config parameter should be added to the docker driver, and possibly documented in duffle so that it actually gets used.

The config parameter should probably just skip this line: https://github.com/deislabs/cnab-go/blob/master/driver/docker/docker.go#L178

astrieanna:
Currently, it is hard to debug the state of an invocation image after
the various files and environment variables have been inserted by the
runtime. By not cleaning up the container, we allow bundle developers
an easier way to debug problems.

The parameter is CLEANUP_CONTAINERS, default is true. I decided to log some feedback if the value is incorrect. This is not what the other settings do, but seems user-friendly.

Fixes #128

cc @youreddy

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -55,11 +55,20 @@ func (d *Driver) Config() map[string]string {
 		"PULL_ALWAYS":         "Always pull image, even if locally available (0|1)",
 		"DOCKER_DRIVER_QUIET": "Make the Docker driver quiet (only print container stdout/stderr)",
 		"OUTPUTS_MOUNT_PATH":  "Absolute path to where Docker driver can create temporary directories to bundle outputs. Defaults to temp dir.",
+		"CLEANUP_CONTAINERS":  "If true, the docker container will be destroyed when it finishes running. If false, it will not be destroyed. The supported values are true and false. Defaults to true.",
 	}
 }
 
 // SetConfig sets Docker driver configuration
 func (d *Driver) SetConfig(settings map[string]string) {
+	// Set default and provide feedback on acceptable input values.
+	value, ok := settings["CLEANUP_CONTAINERS"]
+	if !ok {
+		settings["CLEANUP_CONTAINERS"] = "true"
+	} else if value != "true" && value != "false" {
+		fmt.Printf("CLEANUP_CONTAINERS environment variable has unexpected value %q. Supported values are 'true', 'false', or unset.", value)
+	}
+
 	d.config = settings
 }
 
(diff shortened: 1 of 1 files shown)


driver/docker/docker.go
```diff
@@ -55,11 +55,20 @@ func (d *Driver) Config() map[string]string {
...
 }
 
 // SetConfig sets Docker driver configuration
 func (d *Driver) SetConfig(settings map[string]string) {
+	// Set default and provide feedback on acceptable input values.
+	value, ok := settings["CLEANUP_CONTAINERS"]
+	if !ok {
+		settings["CLEANUP_CONTAINERS"] = "true"
+	} else if value != "true" && value != "false" {
+		fmt.Printf("CLEANUP_CONTAINERS environment variable has unexpected value %q. Supported values are 'true', 'false', or unset.", value)
```
silvin-lubecki: I think the cnab-go runtime shouldn't print something on the stdout without a way to disable it. `SetConfig` should either return an error (would maybe break things), or ignore invalid values and consider them as the default (aka: "true"). WDYT?

driver/docker/docker.go
```diff
@@ -55,11 +55,20 @@ func (d *Driver) Config() map[string]string {
 		"PULL_ALWAYS":         "Always pull image, even if locally available (0|1)",
 		"DOCKER_DRIVER_QUIET": "Make the Docker driver quiet (only print container stdout/stderr)",
 		"OUTPUTS_MOUNT_PATH":  "Absolute path to where Docker driver can create temporary directories to bundle outputs. Defaults to temp dir.",
+		"CLEANUP_CONTAINERS":  "If true, the docker container will be destroyed when it finishes running. If false, it will not be destroyed. The supported values are true and false. Defaults to true.",
```
silvin-lubecki: Maybe we should invert the name and the logic, with something like "KEEP_CONTAINERS" ? So by default it's `false` and I guess the code in `SetConfig` would be simpler?

jeremyrickard: Can you open a follow up issue for that Silvin?

silvin-lubecki: Done here #141

driver/docker/docker.go
astrieanna: I'd prefer to be able to return an error message, but I didn't want to change the interface for all the drivers just to add one option to one driver.

I strongly dislike UIs that involve the user setting an environment variable and then getting zero feedback if their setting is being ignored. If you have to read the code to understand why a feature doesn't seem to be working for you, then that is not a good user experience.

There are a few ways to fix this that I'd be happy with, but they all require changes to the driver interface/contract:
1) `SetConfig` could return an `error`, which would allow UX feedback for invalid config.
2) The drivers could take a stream to write their logs/errors to, and Duffle can decide whether to pass in stdout/stderr or /dev/null based on user config.
3) The drivers could all use the same logging library as Duffle, which should allow Duffle to configure global settings like log-level.

@silvin-lubecki WDYT?

silvin-lubecki: @astrieanna I'd be in favor of modifying `SetConfig` so it can return an error 👍 
But I also think we need a better way to log everything happening in the driver. That's a pain point we have in `docker app`.

PR #140, opened 2019-10-03
chore(packager/import.go): remove unused error, update issue location

vdice:
A bit of cleanup after this logic was imported from deislabs/duffle...

- removes unused custom error
- updates TODO issue URL
- removes unused `Verbose` struct field
- updates the `NewImporter` func signature to just return an `*Importer` object (had previously always returned `nil` for the additional `error` type)

### DIFF
--- a/packager/import.go
+++ b/packager/import.go
@@ -16,7 +16,6 @@ type Importer struct {
 	Source      string
 	Destination string
 	Loader      loader.BundleLoader
-	Verbose     bool
 }
 
 // NewImporter creates a new secure *Importer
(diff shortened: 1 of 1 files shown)


packager/import.go
```diff
@@ -12,38 +11,31 @@ import (
...
-	Verbose     bool
 }
 
 // NewImporter creates a new secure *Importer
 //
 // source is the filesystem path to the archive.
 // destination is the directory to unpack the contents.
 // load is a loader.BundleLoader preconfigured for loading bundles.
-func NewImporter(source, destination string, load loader.BundleLoader, verbose bool) (*Importer, error) {
+func NewImporter(source, destination string, load loader.BundleLoader, verbose bool) *Importer {
```
silvin-lubecki: We don't want to verbose anymore? Should we keep the verbose parameter then?

vdice: Great catch; thank you!  I'll remove.

PR #142, opened 2019-10-07
Rename k8s driver parm from KUBE_CONFIG to KUBECONFIG

glyn:
KUBECONFIG is the de facto standard environment variable for kubectl etc.

Without this change when the user omits KUBE_CONFIG, the default kube config
will be used, which may well point to a valid cluster, though this will
effectively be the wrong cluster when the user has set $KUBECONFIG.

This is a particular issue when kind (https://kind.sigs.k8s.io/) is used as
it doesn't update the default kube config.

### DIFF
--- a/driver/kubernetes/kubernetes.go
+++ b/driver/kubernetes/kubernetes.go
@@ -13,8 +13,6 @@ import (
 	// Convert transitive deps to direct deps so that we can use constraints in our Gopkg.toml
 	_ "github.com/Azure/go-autorest/autorest"
 
-	"github.com/deislabs/cnab-go/bundle"
-	"github.com/deislabs/cnab-go/driver"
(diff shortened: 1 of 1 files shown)


radu-matei:
```
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export vbom.ml/util: unable to deduce repository and source type for "vbom.ml/util": unable to read metadata: unable to fetch raw metadata: failed HTTP request to URL "http://vbom.ml/util?go-get=1": Get http://vbom.ml/util?go-get=1: dial tcp: lookup vbom.ml on 168.63.129.16:53: read udp 10.1.0.4:32977->168.63.129.16:53: i/o timeout
Makefile:41: recipe for target 'bootstrap' failed
```

PR #143, opened 2019-10-08
feat(bundle.go): add DeepCopy for InvocationImage/BaseImage/Image

vdice:
Adds a `DeepCopy` method on a `BaseImage` struct, used also for `DeepCopy` methods for the `InvocationImage` and `Image` structs.


vdice:
(subsequent forced-pushed commit just to rebase w/ master)

Issue #121, opened 2019-09-10
Rely on released code only


carolynvs:
For a very long time now duffle, and now cnab-go has been relying on magic commits in a few packages.

https://github.com/deislabs/cnab-go/blob/2b49a93c113c94ccd6f13ca11509319a16b0a3b1/Gopkg.toml#L13-L28

https://github.com/deislabs/cnab-go/blob/2b49a93c113c94ccd6f13ca11509319a16b0a3b1/Gopkg.toml#L46-L60

This is a problem. It means that while yes, our own library works, for every downstream consumer of the library, they have to figure out what that magic commit is and use an override in _their_ Gopkg.toml to get their code to work properly. 

It also guarantees that as soon as there is a release or change in these dependencies that we apply to our Gopkg.toml, everyone who relies on cnab-go is going to have dep ensure fail when they update. 😅 

It also locks the entire dependency graph onto a single commit, which makes resolution difficult if they are using the package elsewhere their code and need any other version but that commit.

The ideal configuration that we should be striving towards with a 1.0 release is that a consumer of our library can add a constraint for just cnab-go, and they inherit all of our constraints and they _just work_. If we are using any overrides or requires, then that won't work (which is our current state). Moving to modules would change some of the UX problems described here but not the fundamental graph solving problems.

I am not sure why those commits have remained unreleased in those packages for so long? But it's worth figuring out and coming up with a plan to get onto tagged versions.

PR #150, opened 2019-10-19
Clean overridden dependencies

radu-matei:
depends on https://github.com/docker/cnab-to-oci/pull/77
ref #121 

When that PR is merged, we should update this PR, do a release of `cnab-to-oci`, then a release of `cnab-go`.

There are still some dependencies that are still overridden:
```
[[override]]
  name = "github.com/containerd/containerd"
  version = "v1.3.0"

[[override]]
  name = "github.com/docker/distribution"
  #version = "v2.7.1"
  revision = "0d3efadf0154c2b8a4e7b6621fff9809655cc580"

[[override]]
  name = "github.com/docker/docker-credential-helpers"
  version = "v0.6.3"

[[override]]
  name = "golang.org/x/sys"
  revision = "f49334f85ddcf0f08d7fb6dd7363e9e6d6b777eb"
```

- without the `x/sys` override: `github.com/docker/docker/pkg/system/filesys_windows.go:112:24: cannot use uintptr(unsafe.Pointer(&sd[0])) (type uintptr) as type *"github.com/deislabs/cnab-go/vendor/golang.org/x/sys/windows".SECURITY_DESCRIPTOR in assignment`

- `go-containerregistry` - I got this from `duffle` - I think it relates to `image-relocation` - @glyn might be able to help`

- `containerd`, `distribution`, `docker-credential-helpers` - this helps pin compatible versions across `cnab-to-oci`, `cli`, `client`, `docker/docker`.  Feel free to try turning these into constraints rather than overrides, I didn't have any luck so far.

Please test this on all operating systems and supported Go versions - I also added the `osusergo` build tag (see https://github.com/golang/go/issues/23265) - without it, cross compilation on Darwin fails.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,65 +1,20 @@
-[[constraint]]
-  name = "github.com/docker/go"
-  version = "1.5.1-1"
-
-[[constraint]]
-  name = "github.com/stretchr/testify"
-  version = "1.2.2"
-
-[[constraint]]
-  name = "gopkg.in/yaml.v2"
-  version = "2.2.1"
-
-[[constraint]]
-  # version imported by k8s.io/client-go @ kubernetes-1.11.2
-  name = "github.com/Azure/go-autorest"
-  revision = "1ff28809256a84bb6966640ff3d0371af82ccba4"
-
-[[constraint]]
-  name = "github.com/docker/cli"
-  revision = "f95ca8e1ba6c22c9abcdbf65e8dcc39c53958bba"
-
 [[override]]
-  name = "github.com/docker/docker"
-  revision = "f76d6a078d881f410c00e8d900dcdfc2e026c841"
+  name = "github.com/containerd/containerd"
+  version = "v1.3.0"
 
 [[override]]
   name = "github.com/docker/distribution"
-  revision = "83389a148052d74ac602f5f1d62f86ff2f3c4aa5"
+  #version = "v2.7.1"
+  revision = "0d3efadf0154c2b8a4e7b6621fff9809655cc580"
 
 [[override]]
-  name = "k8s.io/apimachinery"
-  revision = "kubernetes-1.11.2"
-
-[[override]]
-  name = "k8s.io/api"
-  revision = "kubernetes-1.11.2"
-
-[[override]]
-  name = "k8s.io/client-go"
-  revision = "kubernetes-1.11.2"
-
-[[override]]
-  name = "k8s.io/kubernetes"
-  revision = "kubernetes-1.11.2"
-
-[[override]]
-  name = "k8s.io/kube-openapi"
-  revision = "d8ea2fe547a448256204cfc68dfee7b26c720acb"
-
-[[override]]
-  name = "github.com/docker/compose-on-kubernetes"
-  revision = "a6086e2369e39c2058a003a7eb42e567ecfd1f03"
-
-[[override]]
-  name = "github.com/containerd/containerd"
-  revision = "7f5f1176dd9fb3cc8d3ce5de91759ed3dc969fa2"
+  name = "github.com/docker/docker-credential-helpers"
+  version = "v0.6.3"
 
 [[override]]
   name = "golang.org/x/sys"
   revision = "f49334f85ddcf0f08d7fb6dd7363e9e6d6b777eb"
 
-[prune]
-  non-go = true
-  unused-packages = true
-  go-tests = true
+[[override]]
+  name = "github.com/google/go-containerregistry"
+  revision = "ef12d49c8daf6a6d72978966d5945e39bb898b4a"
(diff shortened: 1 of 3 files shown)


radu-matei:
Not sure what Azure Pipelines is doing, cancelled the build and reran it, it's green now - https://dev.azure.com/deislabs/cnab-go/_build/results?buildId=3919

radu-matei:
Also, we should probably consider adding the `vendor/` directory to source control, like we do in the rest of the Go projects in CNAB.

Gopkg.toml
```diff
@@ -1,65 +1,20 @@
...
   name = "golang.org/x/sys"
   revision = "f49334f85ddcf0f08d7fb6dd7363e9e6d6b777eb"
 
-[prune]
-  non-go = true
-  unused-packages = true
-  go-tests = true
+[[override]]
+  name = "github.com/google/go-containerregistry"
+  revision = "ef12d49c8daf6a6d72978966d5945e39bb898b4a"
```
vdice: I'm wondering if we can override with a more recent commit for this repo?  Though, perhaps this was pinned at the inherited commit on master?

https://github.com/google/go-containerregistry/commit/0e086aed1d2909810e011e2a073054f2eb4e4df2 brings in a fix, courtesy @glyn, related to supporting pushing images (or indexes) with differing media types, which Porter hit with a use case mentioned in https://github.com/pivotal/image-relocation/issues/34

radu-matei: I would like to defer this change to someone familiar with the expected behavior of this library (cc @glyn).

Gopkg.toml
```diff
@@ -1,65 +1,20 @@
...
-  name = "github.com/containerd/containerd"
-  revision = "7f5f1176dd9fb3cc8d3ce5de91759ed3dc969fa2"
+  name = "github.com/docker/docker-credential-helpers"
+  version = "v0.6.3"
 
 [[override]]
   name = "golang.org/x/sys"
   revision = "f49334f85ddcf0f08d7fb6dd7363e9e6d6b777eb"
 
-[prune]
```
carolynvs: Why were the prune options removed? If you are thinking about checking in vendor eventually, then we should be keeping the prune options.

Gopkg.toml
```diff
@@ -1,65 +1,20 @@
...
-[[constraint]]
-  # version imported by k8s.io/client-go @ kubernetes-1.11.2
-  name = "github.com/Azure/go-autorest"
-  revision = "1ff28809256a84bb6966640ff3d0371af82ccba4"
-
-[[constraint]]
-  name = "github.com/docker/cli"
-  revision = "f95ca8e1ba6c22c9abcdbf65e8dcc39c53958bba"
-
 [[override]]
```
carolynvs: overrides without constraints are not communicated to people who depend on your library. So other people who depend upon cnab-go won't get any help from dep on these dependencies, and won't know that containerd for example should be at version 1.3.0.

Each override needs to be paired with a matching constraint for downstream consumers of the library.

Gopkg.toml
```diff
@@ -1,65 +1,20 @@
...
-  revision = "83389a148052d74ac602f5f1d62f86ff2f3c4aa5"
-
-[[override]]
-  name = "k8s.io/apimachinery"
-  revision = "kubernetes-1.11.2"
-
-[[override]]
-  name = "k8s.io/api"
-  revision = "kubernetes-1.11.2"
+  #version = "v2.7.1"
```
carolynvs: Why are we using that revision instead of v2.7.1? That commented out line is a bit misleading. Whenever we do pin to a specific revision it really helps to explain why in the file, so that people can understand when it's safe to move back to a tag, what issue they should be watching and waiting for it to be merged, etc.

carolynvs:
I agree that a follow-up PR where we commit vendor would be appreciated. 👍

Gopkg.toml
carolynvs: Sorry I forgot to mention the other part of the trick to let you constrain a transitive dependency with dep:

https://github.com/deislabs/porter/blob/master/pkg/dep/dep.go

It's a total hack, sorry you can blame google for killing the project before we got a chance to finish...

Anyhoo, you add a file like that in an out of the way package to get dep to consider it direct dependency. Now you can add constraints, and communicate your dependencies to your downstream consumers properly.

carolynvs: I think you can get away with removing the overrides entirely with that this workaround. Give it a try. 

Then in the downstream consumer's Gokpkg.toml they can reference cnab-go without having to add a bunch of overrides and constraints because cnab-go did all the proper work up-front. That's what porter has been doing with it's mixin repos.

jlegrone:
@radu-matei sorry for the churn, can you rebase this PR now that #161 is merged?

carolynvs:
Closing now that we use go mods.

PR #154, opened 2019-10-28
Add all well known actions

### LINKED ISSUE #153: action.Status doesn't follow the v1.0.0 spec
[`action.Status`](https://github.com/deislabs/cnab-go/blob/master/action/status.go) implements the Run function of `Action` interface with a name define in [`claim.go`](https://github.com/deislabs/cnab-go/blob/ae9303613f21b849bf4edb06adc5b0531b26b1cd/claim/claim.go#L28)

As expected in specification the name should be `io.cnab.status` instead of `status`

glours:
Rename Status action to `io.cnab.status` instead of `status`
Remove Downgrade action

I added all well known actions inside a `well_know.go` file (and remove`status.go`), define all action name as defined in the [specifications](https://github.com/deislabs/cnab-spec/blob/master/804-well-known-custom-actions.md) and implement all the run functions by using the `run-custom` action

Fix #153


action/well_known.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Run executes a status action in an image
+func (i *Status) Run(c *claim.Claim, creds credentials.Set, opCfgs ...OperationConfigFunc) error {
+	return (&RunCustom{Driver: i.Driver, Action: ActionStatus}).Run(c, creds, opCfgs...)
+}
+
+// Run executes a status+JSON action in an image
+func (i *StatusJSON) Run(c *claim.Claim, creds credentials.Set, opCfgs ...OperationConfigFunc) error {
+	return (&RunCustom{Driver: i.Driver, Action: ActionStatusJSON}).Run(c, creds, opCfgs...)
+}
```
jlegrone: I see that this is already in the [spec](https://github.com/deislabs/cnab-spec/blob/master/804-well-known-custom-actions.md), so it may be outside the scope of discussion for this PR. But it seems like there should be no need for a separate `io.cnab.status+json` action when `io.cnab.status` also exists given that `io.cnab.status` could define a [cnab output](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#outputs) for the `status.json` file.

action/well_known.go
```diff
@@ -0,0 +1,66 @@
...
+func (i *DryRun) Run(c *claim.Claim, creds credentials.Set, opCfgs ...OperationConfigFunc) error {
+	return (&RunCustom{Driver: i.Driver, Action: ActionDryRun}).Run(c, creds, opCfgs...)
+}
+
+// Run executes a help action in an image
+func (i *Help) Run(c *claim.Claim, creds credentials.Set, opCfgs ...OperationConfigFunc) error {
+	return (&RunCustom{Driver: i.Driver, Action: ActionHelp}).Run(c, creds, opCfgs...)
+}
+
+// Run executes a status action in an image
```
jlegrone:
```suggestion
// Run executes a log action in an image
```

action/well_known.go
jlegrone: Started discussion in https://github.com/deislabs/cnab-spec/pull/290

jlegrone: Update: the spec change is merged, so let's remove this.

radu-matei:
@glours - apologies for the delay here..
Would you mind rebasing this PR? If not, I'm happy to continue the work on this branch.

glours:
@radu-matei done, the PR will need a another rebase due to test issues on claim status (fix available #186 )

glours:
Updated with the merge of #185 into master

action/well_known.go
```diff
@@ -0,0 +1,60 @@
+package action
+
+import (
+	"github.com/cnabio/cnab-go/claim"
+	"github.com/cnabio/cnab-go/credentials"
+	"github.com/cnabio/cnab-go/driver"
+)
+
+// Well known constants define the Well Know CNAB action to be taken
```
silvin-lubecki: `the Well Known CNAB actions to be taken` instead?

action/well_known.go
```diff
@@ -0,0 +1,60 @@
+package action
```
silvin-lubecki: I think we should avoid underscore in file names and follow the same pattern as packages (see https://stackoverflow.com/a/39625565 )

action/well_known_test.go
```diff
@@ -0,0 +1,7 @@
+package action
+
+// makes sure DryRun, Help, Log, Status, StatusJSON implements Action interface
+var _ Action = &DryRun{}
```
silvin-lubecki: nit: You can use a `var` block here:
```go
var (
    _ Action = &DryRun{}
    _ Action = &Help{}
    _ Action = &Log{}
    _ Action = &Status{}
)
```

PR #157, opened 2019-11-06
Difficulty with transitive dependencies while bumping ggcr

glyn:
This draft PR is evidence of the need for https://github.com/deislabs/cnab-go/issues/156.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -208,19 +208,19 @@
   version = "v3.2.0"
 
 [[projects]]
-  digest = "1:4fc9beef384b8f57f5bb5fa8d5b00efa350c21b8ef85d93f4d24ef9c18be812d"
+  digest = "1:53c2173b1fe2039e40eac73316e0b544e2e8519653a60e4e25a174694e347d6f"
   name = "github.com/docker/cli"
   packages = [
-    "cli",
     "cli/command",
     "cli/config",
     "cli/config/configfile",
     "cli/config/credentials",
+    "cli/config/types",
     "cli/connhelper",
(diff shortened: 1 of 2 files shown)


carolynvs:
I had a conversation with @radu-matei on his other PR https://github.com/deislabs/cnab-go/pull/150/files#r340859692 about how to properly manage dependencies. If you want to manage transitive dependencies with dep, that is how to do it.

glyn:
> I had a conversation with @radu-matei on his other PR https://github.com/deislabs/cnab-go/pull/150/files#r340859692 about how to properly manage dependencies. If you want to manage transitive dependencies with dep, that is how to do it.

Thanks. Although that may be a necessary workaround with dep, it's not ideal because we then have to manually maintain the transitive dependencies and line them up accurately. I think that's reason enough alone to spur us on to go modules, which isn't that painful actually.

squillace:
@glyn you, of course, realize that once you say something's not very painful, you end up being the sole point of responsibility, right? :-)

carolynvs:
I'm just sharing what we can do today, if we aren't ready to jump the 🦈 and switch 15+ repos over to modules in a weekend. I am painfully aware that dep was killed before transitive dependency support was added...

If the group would like to move to modules immediately, I am not going to stand in their way but I will wait for someone else to volunteer. 😀

glyn:
Thanks @carolynvs. When we move to modules, I propose we start on a Monday. Let's plan to keep our weekends for rest. :-)

jlegrone:
@glyn is this still needed now that #161 is merged?

PR #159, opened 2019-11-26
fix(kubernetes): handle long installation names

jlegrone:
Long installation or action names currently cause the Kubernetes driver to fail because of character limits imposed by the platform ([63 chars for label values](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) and [253 chars for resource names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names)).

This PR ensures that only valid DNS-1123 resource names less than the maximum length are generated, and uses annotations instead of labels for storing metadata about which jobs are associated with which operations.

### DIFF
--- a/driver/kubernetes/kubernetes.go
+++ b/driver/kubernetes/kubernetes.go
@@ -3,8 +3,10 @@ package kubernetes
 import (
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -31,6 +33,11 @@ const (
 	k8sContainerName    = "invocation"
 	k8sFileSecretVolume = "files"
 	numBackoffLoops     = 6
+	cnabPrefix          = "cnab.io/"
+)
+
+var (
+	dns1123Reg = regexp.MustCompile(`[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`)
 )
 
 // Driver runs an invocation image in a Kubernetes cluster.
@@ -130,11 +137,14 @@ func (k *Driver) Run(op *driver.Operation) (driver.OperationResult, error) {
 	if k.Namespace == "" {
 		return driver.OperationResult{}, fmt.Errorf("KUBE_NAMESPACE is required")
 	}
-	labelMap := generateLabels(op)
+
 	meta := metav1.ObjectMeta{
 		Namespace:    k.Namespace,
 		GenerateName: generateNameTemplate(op),
-		Labels:       labelMap,
+		Labels: map[string]string{
+			"cnab.io/driver": "kubernetes",
+		},
+		Annotations: generateMergedAnnotations(op, k.Annotations),
 	}
 	// Mount SA token if a non-zero value for ServiceAccountName has been specified
 	mountServiceAccountToken := k.ServiceAccountName != ""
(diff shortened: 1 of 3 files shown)


driver/kubernetes/kubernetes.go
```diff
@@ -363,15 +370,37 @@ func (k *Driver) deleteJob(name string) error {
 }
 
 func generateNameTemplate(op *driver.Operation) string {
-	return fmt.Sprintf("%s-%s-", op.Installation, op.Action)
+	name := fmt.Sprintf("%s-%s", op.Action, op.Installation)
+	if len(name) > 50 {
+		name = name[0:39]
```
glyn: Why does the code check for length greater than 50, but not truncate to the first 50 bytes?

driver/kubernetes/kubernetes.go
```diff
@@ -363,15 +370,37 @@ func (k *Driver) deleteJob(name string) error {
...
+	}
+
+	var result string
+	for _, match := range dns1123Reg.FindAllString(strings.ToLower(name), 10) {
```
glyn: Why 10 occurrences?

driver/kubernetes/kubernetes.go
```diff
@@ -363,15 +370,37 @@ func (k *Driver) deleteJob(name string) error {
 }
 
 func generateNameTemplate(op *driver.Operation) string {
```
glyn: It would be helpful to document what size string this function aims to generate. Also, is it really generating a "name template"??

technosophos:
The current pipeline build failure is:

```
go get github.com/AlekSi/gocov-xml
go get -u github.com/axw/gocov/gocov
go get github.com/jstemmer/go-junit-report
go build ./...
# github.com/docker/cli/cli/config
../../docker/cli/cli/config/config.go:109:18: undefined: os.UserHomeDir
# k8s.io/client-go/transport
../../../k8s.io/client-go/transport/round_trippers.go:70:11: cannot convert klog.V(9) (type klog.Verbose) to type bool
../../../k8s.io/client-go/transport/round_trippers.go:72:11: cannot convert klog.V(8) (type klog.Verbose) to type bool
../../../k8s.io/client-go/transport/round_trippers.go:74:11: cannot convert klog.V(7) (type klog.Verbose) to type bool
../../../k8s.io/client-go/transport/round_trippers.go:76:11: cannot convert klog.V(6) (type klog.Verbose) to type bool
# github.com/pivotal/image-relocation/pkg/image
../../pivotal/image-relocation/pkg/image/name.go:128:55: cannot use digest.dig (type "github.com/opencontainers/go-digest".Digest) as type "github.com/docker/distribution/vendor/github.com/opencontainers/go-digest".Digest in argument to reference.WithDigest
# github.com/google/go-containerregistry/pkg/logs
../../google/go-containerregistry/pkg/logs/logs.go:38:10: l.Writer undefined (type *log.Logger has no field or method Writer)
# github.com/google/go-containerregistry/vendor/github.com/docker/cli/cli/config
../../google/go-containerregistry/vendor/github.com/docker/cli/cli/config/config.go:109:18: undefined: os.UserHomeDir
Makefile:13: recipe for target 'build' failed
make: *** [build] Error 2
##[error]Bash exited with code '2'.
```

Is this a dependency resolution failure in CI that we are not seeing locally?

driver/kubernetes/kubernetes.go
jlegrone: > Also, is it really generating a "name template"??

Yes, in the sense that it's used by `metav1.ObjectMeta.GenerateName`; I've added a comment to clarify this.

driver/kubernetes/kubernetes.go
jlegrone: This was just an arbitrary number since I'm not aware of an API to allow unlimited matches (we're already constrained by the truncation above). Updated to use to max length instead.

driver/kubernetes/kubernetes.go
jlegrone: Good catch, missed this during a refactor.

jlegrone:
> Is this a dependency resolution failure in CI that we are not seeing locally?

Yes; `make bootstrap` and `make integration-test` pass for me locally.

PR #160, opened 2019-11-29
Fix Tests

jlegrone:
CI was failing because of the github org change messing up gopath imports.

We should probably switch to go modules, but in the mean time these import changes will need to be made either way.

Existing developer's checkouts will need to be moved from `$GOPATH/src/github.com/deislabs/cnab-go` to `$GOPATH/src/github.com/cnabio/cnab-go`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,25 +1,26 @@
 We have [good first issues][good-first-issue] for new contributors and [help wanted][help-wanted] issues for our other contributors.
 
-* `good first issue` has extra information to help you make your first contribution.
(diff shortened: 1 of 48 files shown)


vdice:
@jlegrone sorry for the hassle.  When convenient, please rebase off master to kick off CI again and we should be green/ready to merge.  Thanks so much for the fixes here!

vdice:
Tests have passed twice (ex https://brigadecore.github.io/kashti/builds/01dve16aa2xhgwq4d2mtfbpntw) but the notifications to GH are failing due to a bug in Brigade; merging.

PR #164, opened 2019-12-06
fix(*): bump golangci-lint; fix linting err

vdice:
* bumps golangci-lint version to latest
  - this was motivated by seeing quite lengthy output in CI using the previous/outdated version, e.g. https://brigadecore.github.io/kashti/jobs/tests-01dvcyr1x6pk73cm2fh2v6h9xz.  It appears the bump eliminates this issue; see the test run here: https://brigadecore.github.io/kashti/jobs/tests-01dvedzvxfvn6beeb8k1r8a245

* addresses the following linting error seen locally:

```
 $ make lint
golangci-lint run --config ./golangci.yml
driver/kubernetes/kubernetes.go:400:2: S1031: unnecessary nil check around range (gosimple)
	if mergeWith != nil {
	^
make: *** [lint] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ lint:
 	golangci-lint run --config ./golangci.yml
 
 HAS_GOLANGCI     := $(shell $(CHECK) golangci-lint)
-GOLANGCI_VERSION := v1.16.0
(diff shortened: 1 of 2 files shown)


carolynvs:
So say we all

Issue #145, opened 2019-10-11
Bug in image reference handling in Docker Driver


simongdavies:
There is a bug in the way that the docker driver deals with Image references, it ignores any digest value and just use op.Image.Image:

https://github.com/deislabs/cnab-go/blob/26ed63f06aff506ab4bfea63a2a83d341f987580/driver/docker/docker.go#L158-L182

The Kubernetes Driver seems to do this correctly:

https://github.com/deislabs/cnab-go/blob/26ed63f06aff506ab4bfea63a2a83d341f987580/driver/kubernetes/kubernetes.go#L417-L422

Also see #144

silvin-lubecki:
To me `Digest` field is a `Content Digest` you can compute locally, which is different from the `Digested Reference` returned by the registry. It shouldn't be used to fetch an image as a digested reference. `Content Digest` is there to let the runtime check the image, like a check sum.
I think we should rename that field to ContentDigest, like we did in the json annotation.

@radu-matei @glyn please correct me if I'm wrong.

simongdavies:
Interestingly its serialized as contentDigest:

https://github.com/deislabs/cnab-go/blob/26ed63f06aff506ab4bfea63a2a83d341f987580/bundle/bundle.go#L84

Looks like Porter is using it for the Digested Reference. 

Either way one of the implementations seems to be wrong and it would be useful to clarify

PR #166, opened 2019-12-13
Reference images by digest in docker driver

jlegrone:
This builds on the work in https://github.com/cnabio/cnab-go/pull/114 by supporting the same functionality for the docker driver.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -96,6 +96,15 @@ func (i *BaseImage) DeepCopy() *BaseImage {
 	return &i2
 }
 
+// DigestedRef returns an image reference which can be used to pull by digest using Docker or Kubernetes.
+// If no digest is available, only the image will be returned and the boolean value will be false.
+func (i *BaseImage) DigestedRef() (string, bool) {
+	if i.Digest == "" {
+		return i.Image, false
+	}
+	return fmt.Sprintf("%s@%s", i.Image, i.Digest), true
+}
+
 // Image describes a container image in the bundle
 type Image struct {
 	BaseImage   `yaml:",inline"`
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -2,6 +2,7 @@ package bundle
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


bundle/bundle_test.go
```diff
@@ -680,3 +681,63 @@ func TestImageDeepCopy(t *testing.T) {
...
+			},
+		},
+		"tag only": {
+			expected: "foo:bar",
+			image: BaseImage{
+				Image: "foo:bar",
+			},
+		},
+		"digest only": {
+			expected:  "foo@sha256:7ea254518",
```
glyn: I think it would be better to use a full SHA 256 hash to avoid any confusion. Maybe just pad it with zeros?

bundle/bundle.go
```diff
@@ -96,6 +96,15 @@ func (i *BaseImage) DeepCopy() *BaseImage {
 	return &i2
 }
 
+// DigestedRef returns an image reference which can be used to pull by digest using Docker or Kubernetes.
+// If no digest is available, only the image will be returned and the boolean value will be false.
+func (i *BaseImage) DigestedRef() (string, bool) {
+	if i.Digest == "" {
+		return i.Image, false
```
silvin-lubecki: I think returning false here is not accurate, as `i.Image` can be a digested reference itself.  We should test if the image name is a digested reference here and return the result.

bundle/bundle_test.go
```diff
@@ -680,3 +681,63 @@ func TestImageDeepCopy(t *testing.T) {
...
+				Image: "foo",
+			},
+		},
+		"tag only": {
+			expected: "foo:bar",
+			image: BaseImage{
+				Image: "foo:bar",
+			},
+		},
+		"digest only": {
```
silvin-lubecki: We're missing a case here:
```go
"digested reference": {
    expected: "foo@sha256:7ea254518cab82f53938523eab0dc6601a3dd1edf700f1f6235e66e27cbc7986
    hasDigest: true,
    image: BaseImage{
        Image: "foo@sha256:7ea254518cab82f53938523eab0dc6601a3dd1edf700f1f6235e66e27cbc7986",
    },
},
```

silvin-lubecki:
Regarding @glyn comment on this isssue https://github.com/cnabio/cnab-go/issues/145#issuecomment-541554027 I don't think we should use the `contentDigest` field as it is proposed in this PR.

> Regardless of which definition we choose, it seems clear that we can't simply append the contentDigest to the end of the image field to create a digested image reference because the image field may already be digested.

@glyn WDYT?

glyn:
> Regarding @glyn comment on this isssue [#145 (comment)](https://github.com/cnabio/cnab-go/issues/145#issuecomment-541554027) I don't think we should use the `contentDigest` field as it is proposed in this PR.
> 
> > Regardless of which definition we choose, it seems clear that we can't simply append the contentDigest to the end of the image field to create a digested image reference because the image field may already be digested.
> 
> @glyn WDYT?

Of course, I agree with my own comment. But are you saying that the current PR proposes a particular interpretation of `contentDigest`?

silvin-lubecki:
> Of course, I agree with my own comment. But are you saying that the current PR proposes a particular interpretation of contentDigest?

Yes I think here https://github.com/cnabio/cnab-go/pull/166/files#diff-187c9d266554ba3ff27d602db2a12221R105
```go
return fmt.Sprintf("%s@%s", i.Image, i.Digest), true
```
`i.Digest` is the `contentDigest` (this field is unfortunately badly named). I guess it is exactly what your are referring to in your previous comment `we can't simply append the contentDigest to the end of the image field to create a digested image reference`. Or am I totally wrong? 🤔

glyn:
That's correct, but even if the image field is _not_ digested, it only makes sense to add the `contentDigest` on the end if `contentDigest` is a repo digest (and not an image id). We won't know for sure which it is until https://github.com/cnabio/cnab-spec/issues/287 has been decided.

silvin-lubecki:
Agreed, that's why I think we should put that PR on  hold.

bundle/bundle.go
jlegrone: I'm on the fence about this. The spec states that

> If the `contentDigest` is not present, the runtime SHOULD report an error so the user is aware that there is no `contentDigest` provided.

If an image contains what we interpret to be a digest, then some runtimes could certainly opt to use this in lieu of an explicit `contentDigest`. But having the boolean correspond to whether `contentDigest` has actually been set seems most in line with the spec as written.

vdice:
Attempt to clarify the spec (address https://github.com/cnabio/cnab-spec/issues/287) has been issued in https://github.com/cnabio/cnab-spec/pull/384.

Meanwhile, this PR looks great to me.  It looks like there are some unit test suggestions remaining and a rebase is needed.  These changes will come in handy for digest validation in https://github.com/cnabio/cnab-go/pull/227.

vdice:
cnabio/cnab-spec#384 has been merged.

@jlegrone any chance this branch can be rebased when convenient?

PR #169, opened 2019-12-26
Waiting container to stop after started container, not before

fenngwd:
Waiting container to stop is not working in old Docker API version(1.24).

In old Docker API version(1.24), before docker container started, `docker wait` will return immediately. In this case, `ContainerWait` function which creates a goroutine to wait container to stop, may send a message to  `statusc ` before container started, and this will make waiting container to stop fail.

Trying to wait container to stop after started container, and this will make waiting container work well.

Related source code:
[https://github.com/moby/moby/blob/1.12.x/container/state.go#L129](url)

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -231,10 +231,10 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
 		}
 	}()
 
-	statusc, errc := cli.Client().ContainerWait(ctx, resp.ID, container.WaitConditionNextExit)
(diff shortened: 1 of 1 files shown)


silvin-lubecki:
Hello @fenngwd, thank you for this PR. It looks like the DCO bot is complaining, could you signoff your commit?
Also one integration test related to your changes is failing on the CI, can you reproduce it on your side?

fenngwd:
@silvin-lubecki I have signoff my commit, and integration test passed. But it seems check hangs.

fenngwd:
@silvin-lubecki How to solve this? Check hangs...

Issue #172, opened 2020-01-28
Improve support for storing host env externally


carolynvs:
One feature that I have been working on is being able to store a tool's entire host environment, e.g. claims, credential sets, externally instead of on the local host.

I'm going to submit a few PRs around this and wanted to have a single issue to help give context them, as they really all came from a single branch to support this that I split up once I had it all working. 😀 

* Improve crud.Store so it's easier to use the same instance for working with different types of data. As far as crud is concerned they are all `[]byte`, though we do want to group them by type.
* Filesystem is easy and doesn't require connecting or releasing resources but most other data stores need our structs to account for this.
* Make CredentialStrategy generic so that we can lookup credential values from other locations than just the local host. For example, retrieving them from a key/value store like Hashicorp Vault.
* Other smaller related changes.

I'll do my best to make each independent but they may end up stepping on each other since they do affect the same files.

carolynvs:
All the tasks from this issue have been implemented.

PR #173, opened 2020-01-28
Add timestamps to CredentialSet

carolynvs:
This allows tools to display when a credential set was created/modified without relying upon file timestamps which aren't available when credentials are stored in the generic crud datastore.

Since we have this data on Claim, it makes sense to have it on other data that we generally store in the host environment too.

Part of #172 

Signed-off-by: Carolyn Van Slyck <carolyn.vanslyck@microsoft.com>

### DIFF
--- a/credentials/credentialset.go
+++ b/credentials/credentialset.go
@@ -6,9 +6,9 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/cnabio/cnab-go/bundle"
-
 	yaml "gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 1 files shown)


radu-matei:
Would timestamps for both claims and credential sets make sense as part of the spec?

carolynvs:
Credential Sets are non-normative. I'll submit a PR for credential sets adding the timestamps to the examples so they match what a marshaled struct from cnab-go looks like.

Claims are going to be redone in the next week or so to move state and the last result (which includes timestamp) out of the claim. Modified is already part of the spec, but not created. I'll keep an eye out for that and see that its suggested.

carolynvs:
I went to tweak duffle to set these timestamps experimentally and realized that duffle doesn't build against cnab-go anymore. I'll submit a PR in a bit to update to the latest release of cnab-go and switch to go mods.

carolynvs:
I've opened a PR to add timestamps to the credential set spec

https://github.com/cnabio/cnab-spec/pull/318

PR #175, opened 2020-01-28
Update crud.Store to support general storage

carolynvs:
CRUD can be used to work with any type of item, not just claims. Update the interface signature to support specifying what type of item is being stored so that the controlling code, e.g. the claim store or eventually a credential store, can say what type of data is being stored. For example "claims" or "credentials".

By altering the interface, this allows tools to avoid creating multiple instances of the same store to work with a different type of data.

Part of #172 

~🚨 This has some overlap with #174. I tried to make this PR independent but if that PR is merged first, this should be rebased and tweaked first.~ Rebase done.

### DIFF
--- a/utils/crud/mongodb.go
+++ b/utils/crud/mongodb.go
@@ -9,12 +9,12 @@ import (
 )
 
 // MongoClaimsCollection is the name of the claims collection.
-const MongoClaimsCollection = "cnab_claims"
+const MongoCollectionPrefix = "cnab_"
 
 type mongoDBStore struct {
-	session    *mgo.Session
-	collection *mgo.Collection
-	dbName     string
+	session     *mgo.Session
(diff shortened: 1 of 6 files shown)


utils/crud/mongodb.go
```diff
@@ -9,12 +9,12 @@ import (
 )
 
 // MongoClaimsCollection is the name of the claims collection.
-const MongoClaimsCollection = "cnab_claims"
+const MongoCollectionPrefix = "cnab_"
```
vdice: tiny note: we can remove the `claims` specific section from the comment above

carolynvs: I am terrible at updating the comments. 🤦‍♀ Thanks, I'll get that fixed.

PR #176, opened 2020-01-28
Add crud.BackingStore

carolynvs:
BackingStore is a utility for people developing CRUD stores that makes it easier for them to do late binding / initialization when the list/save/delete/read methods are called, instead of in the constructor, when you may not be ready to connect to databases, plugins, etc or return errors.

This also adds a CredentialStore which isn't interesting now but combined with #175, makes more sense.

One question I had is that that ReadAll isn't part of crud.Store but it is very useful for both credential store and claim store, so it's implemented on both structs. And in Porter, we created new composite interfaces with crud.Store to add in ReadAll as well. I'm wondering if missing ReadAll was an oversight or deliberate decision?

Part of #172

~🚨 This has some overlap with ~#174~ and #175. I tried to make this PR independent but if those PRs are merged first, this should be rebased and tweaked first.~ Update: I've rebased #174 and #175


credentials/credstore.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/cnabio/cnab-go/utils/crud"
+)
+
+const ItemType = "credentials"
+
+// ErrNotFound represents a credential set not found in storage
+var ErrNotFound = errors.New("Credential set does not exist")
+
+// Store is a persistent store for
```
vdice: `...for credential sets`?

carolynvs: for puppies obvs 🐶

credentials/credstore.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/cnabio/cnab-go/utils/crud"
+)
+
+const ItemType = "credentials"
```
radu-matei: Who is supposed to use `ItemType`?
Also nit: could you add a comment for it?

utils/crud/backingStore.go
```diff
@@ -0,0 +1,119 @@
+package crud
```
radu-matei: nit (that you can absolutely ignore): the Go file name for the claim store is `claimstore.go`, while this is `backingStore.go`. Would it make sense ti change either of them to be consistent?

claim/claimstore.go
```diff
@@ -52,22 +53,24 @@ func (s Store) Read(name string) (Claim, error) {
...
+	results, err := s.backingStore.ReadAll()
 	if err != nil {
-		return claims, err
+		return nil, err
 	}
 
-	for _, c := range list {
-		cl, err := s.Read(c)
+	claims := make([]Claim, len(results))
+	for i := range claims {
```
silvin-lubecki: Totally nit, but what about ranging on results instead of claims?
```go
for i, bytes := range results{
    var claim Claim
    if err := json.Unmarshal(bytes, &claim); err != nil {
        return nil, fmt.Errorf("error unmarshaling claim: %v", err)
    }
    claims[i] = claim
}
```

utils/crud/backingStore.go
```diff
@@ -0,0 +1,119 @@
...
+type BackingStore struct {
+	AutoClose    bool
+	closed       bool
+	backingStore Store
+}
+
+func NewBackingStore(store Store) *BackingStore {
+	return &BackingStore{
+		AutoClose:    true,
+		closed:       true,
```
silvin-lubecki: nit: What about renaming `closed` to `opened` ? I think it fits better here in the "constructor":
```go
opened: false,
```
And also below in the `Connect` function, I think we should avoid double negations:
```go
if s.opened{
    return nil
}
```
WDYT?

utils/crud/backingStore.go
```diff
@@ -0,0 +1,119 @@
+package crud
+
+var _ Store = &BackingStore{}
+
+// BackingStore wraps another store that may have Connect/Close methods that
+// need to be called.
+// - Connect is called before a method when the connection is closed.
+// - Close is called after each method when AutoClose is true (default).
+type BackingStore struct {
+	AutoClose    bool
```
silvin-lubecki: I think `AutoClose` shouldn't be exposed, and should be configured only once at construction time. We can either add a boolean to the `NewBackingStore` function, or have a functional options pattern (but here it may be too much as we only have one parameter? 🤔 ). WDYT?

utils/crud/backingStore.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// ReadAll retrieves all the items.
+func (s *BackingStore) ReadAll() ([][]byte, error) {
+	if s.AutoClose {
+		defer s.Close()
+	}
+
+	autoClose := s.AutoClose
+	s.AutoClose = false
```
silvin-lubecki: What about deferring a function which restores the `AutoClose` value? Here as we mutate `s` and can early return if an error occurs, we will never restore the original `AutoClose` value.

utils/crud/backingStore.go
```diff
@@ -0,0 +1,119 @@
...
+	return nil
+}
+
+func (s *BackingStore) List() ([]string, error) {
+	err := s.Connect()
+	if err != nil {
+		return nil, err
+	}
+
+	if s.AutoClose {
```
silvin-lubecki: nit: What about extracting this to a small helper function?
```go
func (s *BackingStore) autoClose() {
    if !s.closed && s.AutoClose {
        s.Close()
    }
}

func (s *BackingStore) List() ([]string, error) {
    defer s.autoClose()
    ...
}

func (s *BackingStore) Save(name string, data []byte) error {
    defer s.autoClose()
    ...
}
...
```

credentials/credstore.go
carolynvs: `ItemType` is intended to be used only by `CredentialStore` but I exported it in case it's useful for others (probably for tests) so that they aren't making their own hard-coded strings for "credentials".

I'll add a comment.

utils/crud/backingStore.go
carolynvs: Yes, I think that is an artifact of me waffling around while working on a solution. Let me fix that.

utils/crud/backingStore.go
carolynvs: I was thinking about that, works better too because the default is false without having to set it in a constructor. I'll update it.

credentials/credstore.go
carolynvs: Oh! I realize now why you asked about this. 🤦‍♀  This is a line that belongs to the other PR #175. Let me remove it from this PR.

utils/crud/backingStore.go
carolynvs: I'll switch to use `backingstore` to be consistent with other files in this repo

utils/crud/backingStore.go
carolynvs: I made it an exported field because I didn't expect anyone to use it often honestly, and wanted AutoClose to be defaulted to True. This seemed the most straightforward way to let people change that if they needed more control.

Is there a problem with the exported field that I am overlooking? Just to limit exports?

carolynvs:
I've add commits to fixup the PR based on review feedback and also rebased the PR on master and resolved the conflicts.

PR #181, opened 2020-02-20
ref(*): change Name to Installation on Claim

vdice:
* change `Name` to `Installation` on a `Claim`

Ref https://github.com/cnabio/cnab-spec/pull/310


claim/claim.go
```diff
@@ -35,13 +35,13 @@ const (
...
 // a CNAB package.
 type Claim struct {
-	Name       string                 `json:"name"`
-	Revision   string                 `json:"revision"`
-	Created    time.Time              `json:"created"`
-	Modified   time.Time              `json:"modified"`
-	Bundle     *bundle.Bundle         `json:"bundle"`
-	Result     Result                 `json:"result,omitempty"`
-	Parameters map[string]interface{} `json:"parameters,omitempty"`
+	Installation string                 `json:"installation"`
```
carolynvs: We put the spec version in bundle.json so tools can tell what version of the spec it has been serialized to match. 

Seeing a breaking change like this and thinking through how tools can detect and migrate existing claims gracefully while also supporting the new claim format, I think we should also start putting the schema version for the claims spec on the `Claim` struct as well.

That way a tool can deserialize to a map or scan the json document for schema version, figure out if it's the old one with `name` or the new one with `installation`, handle the migration if needed, then read it into the new cnab-go struct.

I'll make a new issue in cnab-go for that.

claim/claim_test.go
```diff
@@ -132,7 +132,7 @@ func TestClaimSchema(t *testing.T) {
 	claimBytes, err := json.Marshal(exampleClaim)
 	assert.NoError(t, err, "failed to json.Marshal the claim")
 
-	url := "https://raw.githubusercontent.com/cnabio/cnab-spec/master/schema/claim.schema.json"
+	url := "https://cnab.io/v1/claim.schema.json"
```
carolynvs: 🎉

claim/claim.go
carolynvs: Ok, I created #182

vdice:
@carolynvs are you good with this PR as it is?  Alas, as mentioned, this breaking change doesn't really prompt a claim version change from `v1.0.0-WD` since it's still in the working draft phase...

carolynvs:
Yup! Sorry for the confusion, that's why I created the follow-up issue. 👍

Issue #182, opened 2020-02-21
Add schema version to the claim struct


carolynvs:
Just like we put the schema version for the CNAB core spec on the bundle.json, so that tools can detect the format of the data contained, we need to do the same for claims. Especially now that we are introducing breaking changes that tools will need to detect and migrate gracefully.

See https://github.com/cnabio/cnab-go/pull/181/files#r382648846, the PR where we changed Claim.Name to Claim.Installation for context.

PR #183, opened 2020-02-21
feat(*): create/use SchemaVersion for Claims, Bundles

vdice:
* Creates a `SchemaVersion` alias
* Updates Bundle and Claim objects to use this type

Closes https://github.com/cnabio/cnab-go/issues/182


bundle/bundle.go
```diff
@@ -184,9 +184,9 @@ func ValuesOrDefaults(vals map[string]interface{}, b *Bundle) (map[string]interf
 
 // Validate the bundle contents.
 func (b Bundle) Validate() error {
-	_, err := semver.NewVersion(b.SchemaVersion)
+	err := b.SchemaVersion.Validate()
 	if err != nil {
-		return fmt.Errorf("invalid bundle schema version %q: %v", b.SchemaVersion, err)
+		return pkgErrors.Wrapf(err, "bundle validation failed")
```
carolynvs: Every time we switch from `errors` to `pkg/errors` an angel gets its wings 🙇‍♀

claim/claim.go
```diff
@@ -7,10 +7,16 @@ import (
...
 	"github.com/oklog/ulid"
+	"github.com/pkg/errors"
 
 	"github.com/cnabio/cnab-go/bundle"
+	"github.com/cnabio/cnab-go/utils/schemaversion"
 )
 
+// DefaultSchemaVersion represents the schema version of the Claim
+// that this library returns by default
+const DefaultSchemaVersion = schemaversion.SchemaVersion("v1.0.0-WD")
```
carolynvs: hmm, gosh I just realized that since we are still in working draft there's no way to detect breaking changes since the version stays the same until we get out of WD. 🤦‍♀ This version schema is so very unhelpful...

I was originally expecting that we could bump this to v1.0.1-WD but I think since we never released 1.0 that isn't right. We should ask around.

To be clear, we still need this change, but it may not help for this particular breaking change due to the weak versioning protocol...

PR #188, opened 2020-03-02
ref(*): add schema validation

vdice:
* fetch needed CNAB schemas at pinned version(s) for test/runtime validation
* use for bundle and claim schema validation
* update bundle schemaVersion to v1.0.0

Closes https://github.com/cnabio/cnab-go/issues/194


radu-matei:
I don't have a strong preference here - but would it make sense to have a top-level `schema` package?

Clearly this would be a breaking change for tools using the current `utils` package, so we should keep this in mind.

utils/schemavalidation/schemavalidation.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/qri-io/jsonschema"
+)
+
+// Validate validates the provided objectBytes against the provided CNAB-Spec schemaType
+// The schemaType must be a valid schema currently hosted at https://cnab.io/v1/*.schema.json
+func Validate(schemaType string, objectBytes []byte) ([]jsonschema.ValError, error) {
+	url := fmt.Sprintf("https://cnab.io/v1/%s.schema.json", schemaType)
+	req, err := http.NewRequest("GET", url, nil)
```
silvin-lubecki: I don't know if it's a good thing to make that http request upon validation, as it will fail in airgap environment.
And I also have some concerns with tests depending on external resources (so it might break all of a sudden without knowing why). Maybe we should "vendor" those files in cnab-go ?

radu-matei: That's a good point.
We should ensure we keep the schemas in sync with whatever CNAB version we're implementing here, and it may be a great opportunity to add tests that validate all bundles we produce against the local schemas as well (not in this PR though).

vdice: Indeed, we'll need to look for a more robust solution.  I'll mark this PR as WIP for now.  I've started a convo in the CNAB Slack channel with ideas...

vdice:
What do y'all think of the current approach?  Added a Makefile target to fetch needed schemas of a given version and then consumes these via packr for validation at runtime.  I originally intended to `.gitignore` the schemas fetched from remote but then assumed they actually needed to be committed to source for downstream/consumers of this library... is that accurate?

utils/schemavalidation/schemavalidation.go
vdice: Updated this PR with an approach that injects the spec versions at build-time in addition to fetching the actual schema files and baking them in.. PTAL when convenient.

utils/schemaversion/schemaversion.go
```diff
@@ -20,3 +21,12 @@ func (v SchemaVersion) Validate() error {
...
 	return nil
 }
+
+// GetSemverSchemaVersion returns a SchemaVersion from the provided string,
+// trimming the non-semver prefix according to schema versioning in the
+// cnabio/cnab-spec repo
+func GetSemverSchemaVersion(schemaVersion string) SchemaVersion {
+	r := regexp.MustCompile("^cnab-[a-z]+-(.*)")
+	match := r.FindStringSubmatch(schemaVersion)
+	return SchemaVersion(match[1])
```
silvin-lubecki: `match` can be an empty/nil slice if `schemaVersion` doesn't match the regexp, so I guess this function should return an error instead?

utils/schemavalidation/schemavalidation.go
```diff
@@ -0,0 +1,43 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/gobuffalo/packr/v2"
+	"github.com/pkg/errors"
+	"github.com/qri-io/jsonschema"
+)
+
+// NewSchemaBox returns a *packer.Box with the schema files from the schema sub-directory
+func NewSchemaBox() *packr.Box {
```
silvin-lubecki: nit: I think this function doesn't need to be exported 🤔

utils/schemavalidation/schemavalidation.go
```diff
@@ -0,0 +1,43 @@
...
+func Validate(schemaType string, objectBytes []byte) ([]jsonschema.ValError, error) {
+	schemaData, err := NewSchemaBox().Find(fmt.Sprintf("%s.schema.json", schemaType))
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read the schema data for type %q", schemaType)
+	}
+
+	rs := &jsonschema.RootSchema{}
+	err = json.Unmarshal(schemaData, rs)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to json.Unmarshal %s schema", schemaType)
```
silvin-lubecki: nit: `"failed to unmarshal %q schema"` ?

utils/schemaversion/schemaversion.go
silvin-lubecki: Ok just found that it's used to initialize a var:
https://github.com/cnabio/cnab-go/pull/188/files#diff-36a2eeedaad2da2e03f68eb1036b88c2R23
Maybe could we move it to an `init()` function and handle the error here (even if I don't like `init` 😅 )?
WDYT?

bundle/bundle.go
```diff
@@ -14,6 +14,15 @@ import (
 	pkgErrors "github.com/pkg/errors"
 )
 
+// CNABSpecVersion represents the CNAB Spec version of the Bundle
+// Schema injected at build-time.
```
silvin-lubecki: So if I use cnab-go in my project, I need to add build flags to get the right version?
Why don't we provide a default value, as this code base is targeting a specific version ?

vdice: Hmm, good point.  In the latest changeset, I've added defaults.

I may explore updating the Makefile logic around fetching schema at build-time to run via a go program such that the canonical default versions are only in one place.

utils/schemaversion/schemaversion.go
vdice: I refactored to add `GetDefaultSchemaVersion()` methods for both Bundles and Claims to replace the global variable approach.  WDYT?

bundle/bundle.go
vdice: Ok, added the go program... now the schema versions are set in one place (library code) and the program ingests these to determine which remote schemas to pull at build-time.

vdice:
@silvin-lubecki this is ready for re-review.  PTAL when convenient.  Thanks!

radu-matei:
Local testing seems to fail for me:

```
--- FAIL: TestClaimSchema (0.46s)
    claim_test.go:154: claim validation against the JSON schema failed:
    claim_test.go:156: /result: {"action":"install",... "modified" value is required
    claim_test.go:156: /: {"bundle":{"descript... "action" value is required
```

utils/schemavalidation/schemavalidation.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/qri-io/jsonschema"
+)
+
+// newSchemaBox returns a *packer.Box with the schema files from the schema sub-directory
+func newSchemaBox() *packr.Box {
+	return packr.New("github.com/cnabio/cnab-go/utils/schemavalidation/schema", "./schema")
+}
+
+// Validate validates the provided objectBytes against the provided CNAB-Spec schemaType
+func Validate(schemaType string, objectBytes []byte) ([]jsonschema.ValError, error) {
```
jlegrone, line 20: I would rather expose a separate method for bundle, claim, etc. schema validation, as this presents a clearer API to end users.

It would also be nice if we had convenience methods that allow passing the native go types rather than raw bytes, as this would provide type checking at compile time (though admittedly at the cost of re-serializing data twice in some cases).

utils/schemavalidation/schemavalidation.go
```diff
@@ -0,0 +1,43 @@
+//go:generate packr2
+
+package schemavalidation
```
jlegrone, line 3: I'm on board with shortening this package name to `schema` and moving out of utils. External consumers would then call functions like `schema.ValidateBundle`.

jlegrone: See also https://golang.org/doc/effective_go.html#package-names

> By convention, packages are given lower case, single-word names

utils/schemavalidation/schemavalidation.go
vdice: Good call.  Updated to add `ValidateBundle`, `ValidateClaim`.  Follow-ups to add convenience methods sounds great if you'd like to take that on.

claim/claim_test.go
```diff
@@ -195,33 +193,16 @@ func TestMarshal_AllFields(t *testing.T) {
 }
 
 func TestClaimSchema(t *testing.T) {
-	t.Skip("This test is pending an alternate, offline-friendly implementation; see https://github.com/cnabio/cnab-go/issues/194")
+	t.Skip("This test is pending non-trivial updates to the Claim and Result objects: https://github.com/cnabio/cnab-go/issues/202")
```
vdice, line 196: Note this test is currently still being skipped.  Quite a few changes to the Claim and Result objects are coming down the pike and will need a follow-up to update and start running this test.

vdice:
@jlegrone @radu-matei @silvin-lubecki PTAL when convenient.  Will be good to get this merged to iterate on, especially for forthcoming changes per the Claim spec.

action/action_test.go
```diff
@@ -40,7 +40,7 @@ var mockSet = credentials.Set{
 func newClaim() *claim.Claim {
 	now := time.Now()
 	return &claim.Claim{
-		SchemaVersion: claim.DefaultSchemaVersion,
+		SchemaVersion: "1.0.0-DRAFT+d7ffba8",
```
carolynvs, line 43: Apologies if this has been asked already and I am just missing it but is there a reason why you aren't using `claim.GetDefaultSchemaVersion()` here instead of copying the string value in a few places throughout the tests?

schema/validation.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, errors.Wrapf(err, "failed to read the schema data for type %q", schemaType)
+	}
+
+	rs := &jsonschema.RootSchema{}
+	err = json.Unmarshal(schemaData, rs)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to json.Unmarshal %q schema", schemaType)
+	}
+
+	err = rs.FetchRemoteReferences()
```
carolynvs, line 42: Again still catching up but is this PR supposed to address airgapped scenarios or was that put off for a follow-up? If not, won't we need to also embed https://github.com/cnabio/cnab-spec/blob/master/schema/definitions.schema.json as well so that we have all of the definitions embedded locally and don't need to fetch anything at runtime?

schema/version.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return fmt.Errorf("invalid schema version %q: %v", version, err)
+	}
+	return nil
+}
+
+// GetSemver returns a proper semver Version from the provided string,
+// trimming the non-semver prefix according to schema versioning protocol in the
+// cnabio/cnab-spec repo
+func GetSemver(schemaVersion string) (Version, error) {
```
carolynvs, line 28: Would you please add a test for this function?

action/action_test.go
vdice: Not sure my train of thought; nevertheless, updated to use that method instead of hard-coding.

schema/validation.go
vdice: Great point.  I've updated in https://github.com/cnabio/cnab-go/pull/188/commits/3783131f4fde52d874e8ea48bd36ff8871496a17 wherein I used a different json schema library for validation after fighting with the qri-io library used elsewhere to do the same... WDYT?

schema/version.go
vdice: Added, which uncovered a bit of lacking functionality in the method under test -- thank you!

schema/validation.go
```diff
@@ -3,51 +3,78 @@
...
 
 	"github.com/gobuffalo/packr/v2"
 	"github.com/pkg/errors"
-	"github.com/qri-io/jsonschema"
+
+	// Using this library as qri-io/jsonschema doesn't appear to have
+	// first-class support for adding auxiliary/ref'd sub-schemas,
+	// apart from fetching remote references over the network
+	// (which doesn't support airgapped scenarios)
+	"github.com/xeipuuv/gojsonschema"
```
carolynvs, line 15: 👍

schema/version_test.go
```diff
@@ -35,3 +35,43 @@ func TestValidate(t *testing.T) {
...
+		expected Version
+		err      string
+	}{{
+		name:     "empty",
+		version:  "",
+		expected: Version(""),
+		err:      `no semver submatch for schemaVersion "" using regex "^cnab-[a-z]+-(.*)"`,
+	}, {
+		name:     "no match",
+		version:  "cnabby-core-1.0.0",
```
carolynvs, line 52: 🦀 forever!

PR #190, opened 2020-03-02
feat(claim): add Result validation

vdice:
* adds validation for the `Result` object, including its `Status` property, according to the [Claims Spec](https://github.com/cnabio/cnab-spec/blob/master/400-claims.md)

Note for reviewers:  Do we want to modify the function signature for the [Update](https://github.com/cnabio/cnab-go/blob/master/claim/claim.go#L85) method on a `Claim` to return an `error`?  We can then check to be sure the supplied `action` and `status` form a valid `Result`; if not, we would return an error.

### DIFF
--- a/claim/claim.go
+++ b/claim/claim.go
@@ -48,9 +48,8 @@ type Claim struct {
 	BundleReference string                      `json:"bundleReference,omitempty"`
 	Result          Result                      `json:"result,omitempty"`
 	Parameters      map[string]interface{}      `json:"parameters,omitempty"`
-	// Outputs is a map from the names of outputs (defined in the bundle) to the contents of the files.
-	Outputs map[string]interface{} `json:"outputs,omitempty"`
-	Custom  interface{}            `json:"custom,omitempty"`
+	Outputs         map[string]interface{}      `json:"outputs,omitempty"`
+	Custom          interface{}                 `json:"custom,omitempty"`
 }
 
 // ValidName is a regular expression that indicates whether a name is a valid claim name.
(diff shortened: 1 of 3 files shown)


radu-matei:
Yes, I think `Update` should indeed return an `error`.

vdice:
> Yes, I think Update should indeed return an error.

@radu-matei I was also thinking about this afterwards... while doing so, should we adjust the arg set from `action, status string` to `res Result`?

radu-matei:
Not sure about that - interestingly enough, we never update the `Message` field from a `Result`, so we should at least add that as well.

The question with respect to passing a `Result` is whether the caller has the object constructed beforehand, or it needs to construct it to call `Update`.
Either way, I don't have a strong preference.

claim/claim.go
```diff
@@ -105,9 +118,16 @@ func ULID() string {
...
 // Validate the Claim
 func (c Claim) Validate() error {
+	// validate the schemaVersion
 	err := c.SchemaVersion.Validate()
 	if err != nil {
 		return errors.Wrapf(err, "claim validation failed")
 	}
+
+	// validate the result
+	err = c.Result.Validate()
```
silvin-lubecki: nit: I think we can simplify with:
```go
return errors.Wrapf(c.Result.Validate(), "claim validation failed")
```

silvin-lubecki:
This PR needs a rebase 🐯

vdice:
Rebased.  Note, however, that the test of a claim against the current claim schema from https://cnab.io will fail.  @carolynvs will be working on changes to the Claim/Result structures per the updated schema.

vdice:
I've pushed a commit to skip the claim schema test for now.  Issue to re-enable (and add the rest of the implementations under test) is being tracked in https://github.com/cnabio/cnab-go/issues/194

claim/claim.go
vdice: Updated; thank you.

vdice:
@radu-matei I am going to hold off on making any changes to the `Update` method as I am predicting it will change further with recent amendments to the spec (splitting claim from result, etc.)

PR #191, opened 2020-03-03
feat(bundle): add param and cred validation

vdice:
* adds Parameter and Credential validation, per spec
* validates both when validating a Bundle
* moved a few other validation functions closer to the structs they validate

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -119,6 +119,16 @@ func (img *InvocationImage) DeepCopy() *InvocationImage {
 	return &img2
 }
 
+// Validate the image contents.
+func (img InvocationImage) Validate() error {
+	switch img.ImageType {
+	case "docker", "oci":
+		return validateDockerish(img.Image)
+	default:
+		return nil
+	}
+}
+
 // Location provides the location where a value should be written in
 // the invocation image.
 //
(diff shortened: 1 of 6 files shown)


bundle/bundle.go
```diff
@@ -119,6 +119,16 @@ func (img *InvocationImage) DeepCopy() *InvocationImage {
 	return &img2
 }
 
+// Validate the image contents.
+func (img InvocationImage) Validate() error {
+	switch img.ImageType {
+	case "docker", "oci":
+		return validateDockerish(img.Image)
```
silvin-lubecki: Note that as we already vendor `docker/distribution`, we can use https://github.com/docker/distribution/blob/master/reference/reference.go#L189 instead to parse image reference?

bundle/bundle.go
```diff
@@ -128,6 +138,15 @@ type Location struct {
 	EnvironmentVariable string `json:"env,omitempty" yaml:"env,omitempty"`
 }
 
+// Validate the Location
+func (l Location) Validate() error {
+	forbiddenPath := "/cnab/app/outputs"
+	if strings.HasPrefix(l.Path, forbiddenPath) {
+		return fmt.Errorf("Path must not be a subpath of %q", forbiddenPath)
```
silvin-lubecki: Maybe add the actual `path` to the error message? And I think error message shouldn't start with an upper case, as they can be wrapped.

bundle/bundle.go
radu-matei: I'm not sure whether this should necessarily be part of this PR, but I agree we should have more robust validation.

silvin-lubecki: I can totally be in a follow up, as `validateDockerish` wasn't introduced in this PR.

bundle/bundle.go
vdice: Good call; added.

bundle/bundle.go
vdice: Good call; filed https://github.com/cnabio/cnab-go/issues/196

vdice:
Updated/rebased; PTAL when convenient @silvin-lubecki .

PR #193, opened 2020-03-03
feat(action): add cnab/claim.json to op files

vdice:
* Add the claim to the operation file list via `/cnab/claim.json` per https://github.com/cnabio/cnab-spec/blob/master/400-claims.md#files

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -194,7 +194,6 @@ func opFromClaim(action string, stateless bool, c *claim.Claim, ii bundle.Invoca
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal bundle contents: %s", err)
 	}
-
 	files["/cnab/bundle.json"] = string(bundleBytes)
 
 	imgMap, err := getImageMap(c.Bundle)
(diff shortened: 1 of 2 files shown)


action/action.go
```diff
@@ -203,6 +202,12 @@ func opFromClaim(action string, stateless bool, c *claim.Claim, ii bundle.Invoca
 	}
 	files["/cnab/app/image-map.json"] = string(imgMap)
 
+	claimBytes, err := json.Marshal(c)
```
silvin-lubecki: nit: in the claim store we use `MarshalIndent` [instead](https://github.com/cnabio/cnab-go/blob/master/claim/claimstore.go#L38). Maybe we could factorize the code and add it as a claim function, like we do on [bundles](https://github.com/cnabio/cnab-go/blob/master/bundle/bundle.go#L64) ?

silvin-lubecki:
This PR needs a rebase 🦁

action/action.go
vdice: Created https://github.com/cnabio/cnab-go/issues/195 to track this item

PR #197, opened 2020-03-04
Sync claim status names with the spec

carolynvs:
Synchronize the claim status names with recent spec changes from

* https://github.com/cnabio/cnab-spec/pull/344
    * success -> succeeded
    * failure -> failed
* https://github.com/cnabio/cnab-spec/pull/345
  * +cancelled
* https://github.com/cnabio/cnab-spec/pull/340
  * +running


carolynvs:
This is in draft until all the PRs are merged.

claim/claim.go
```diff
@@ -19,10 +19,12 @@ const DefaultSchemaVersion = schemaversion.SchemaVersion("v1.0.0-WD")
...
 // Status constants define the CNAB status fields on a Result.
 const (
-	StatusSuccess = "success"
-	StatusFailure = "failure"
-	StatusPending = "pending"
-	StatusUnknown = "unknown"
+	StatusSuccess  = "succeeded"
+	StatusCanceled = "canceled"
+	StatusRunning  = "running"
+	StatusFailure  = "failed"
```
radu-matei: This might be controversial - but do we get any value from also updating the variable name as well - e.g. `StatusFailed`?

carolynvs: I didn't suggest it because it would require even more code changes for consumers but with very little reward. 😀 But now is the time to make breaking changes so if we do want it to match exactly, I can make that change?

silvin-lubecki: I vote for updating the names but we can also add compatibility constants, documented as depecated:
```go
const {
	StatusSucceeded  = "succeeded"
	StatusCanceled = "canceled"
	StatusFailed  = "failed"
	StatusRunning  = "running"
	StatusPending  = "pending"
	StatusUnknown  = "unknown"
        // DEPRECATED
	StatusSuccess = StatusSucceeded
	StatusFailure = StatusFailed
}
```

carolynvs:
I've updated it to have both the deprecated status names and the new status names that match (StatusFailed/StatusSucceeded), and marked this ready for review now that all PRs in the spec have merged.

PR #199, opened 2020-03-15
Allow docker driver to run in privileged mode

jdolitsky:
Use case: build and push my containers as steps in a "porter install"

In order to build images within the context of a "docker run", the --privileged flag appears to be required, even when using daemonless tools such as [img](https://github.com/genuinetools/img).


jdolitsky:
Closing for now, seems that everything that needs to be accomplished can be done via `dockerDriver.AddConfigurationOptions`

See https://github.com/deislabs/porter/pull/951 for more info

PR #201, opened 2020-04-07
Improve matching of not found errors 

carolynvs:
* **Export crud.MockStore and allow mocking errors**
    The MockStore struct is very useful for tests outside of the crud package, since it has an in-memory implementation of the crud.Store interface. This adds the ability to mock errors to test out how code that wrap the store handle edge cases and then exports the struct.
* **Improve matching of not found errors**
    Previously in the claim and credential store, we matched exactly on the underlying not found errors from the crud store. This doesn't work well when the store may be being wrapped in layers, for example in Porter's case the store error is wrapped in an net/rpc error so the error doesn't match exactly.

    This changes the matching to allow for how errors are usually wrapped, with an error message appended and the original message preserved within the error message.

    We care because it allows the claim store and credential store to detect "not found" scenarios and present a more clear error message such as "Claim not found" instead of "file not found" when in many cases the claim may not be represented as a file anyway.

### DIFF
--- a/claim/claimstore.go
+++ b/claim/claimstore.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/cnabio/cnab-go/utils/crud"
 )
@@ -46,7 +47,7 @@ func (s Store) Save(claim Claim) error {
 func (s Store) Read(name string) (Claim, error) {
 	bytes, err := s.backingStore.Read(ItemType, name)
 	if err != nil {
-		if err == crud.ErrRecordDoesNotExist {
+		if strings.Contains(err.Error(), crud.ErrRecordDoesNotExist.Error()) {
 			return Claim{}, ErrClaimNotFound
 		}
 		return Claim{}, err
--- a/utils/crud/mock_store.go
+++ b/utils/crud/mock_store.go
@@ -3,9 +3,6 @@ package crud
 import (
 	"fmt"
 	"strconv"
-	"testing"
-
-	"github.com/stretchr/testify/assert"
 )
 
 // The main point of these tests is to catch any case where the interface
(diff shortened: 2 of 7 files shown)


claim/claimstore.go
```diff
@@ -46,7 +47,7 @@ func (s Store) Save(claim Claim) error {
 func (s Store) Read(name string) (Claim, error) {
 	bytes, err := s.backingStore.Read(ItemType, name)
 	if err != nil {
-		if err == crud.ErrRecordDoesNotExist {
+		if strings.Contains(err.Error(), crud.ErrRecordDoesNotExist.Error()) {
```
carolynvs, line 50: I am not using the errors.Cause function because not all errors are coming from that package. Unfortunately for example when errors are bubbled up from net/rpc (which plugins use in Porter), they get up being special rpc error struct and I need basic string matching instead of exact equality on the error struct to make this work.

utils/crud/mock_store.go
```diff
@@ -3,46 +3,49 @@ package crud
...
 // The main point of these tests is to catch any case where the interface
 // changes. But we also provide a mock for testing.
 var _ Store = &MockStore{}
 
 const (
-	ConnectCount  = "connect-count"
-	CloseCount    = "close-count"
-	TestItemType  = "test-items"
-	MockStoreType = "mock-store"
+	connectCount  = "connect-count"
```
carolynvs, line 13: These previously were not really exported because they were in a test file. I'm making sure that they stay unexported since they are specific to testing the mock store and backing store but I don't anticipate that they will be useful to people who want to use the MockStore.

utils/crud/mock_store_test.go
```diff
@@ -0,0 +1,19 @@
+package crud
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMockStore(t *testing.T) {
```
silvin-lubecki, line 9: I think it's the first time I saw a mock tested 😸

carolynvs: This test has been around since the beginning I think? TRUST NO ONE (including our mocks) 😉

PR #203, opened 2020-04-10
feat(docker.go): access Docker configuration options

vdice:
* Add ability to access the container/host config from a Docker driver

This is motivated by the desire to unit test in this area for runtimes that modify these configurations.

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -31,6 +31,8 @@ type Driver struct {
 	dockerConfigurationOptions []ConfigurationOption
 	containerOut               io.Writer
 	containerErr               io.Writer
+	containerHostCfg           *container.HostConfig
+	containerCfg               *container.Config
 }
 
 // Run executes the Docker driver
@@ -46,6 +48,12 @@ func (d *Driver) Handles(dt string) bool {
 // AddConfigurationOptions adds configuration callbacks to the driver
 func (d *Driver) AddConfigurationOptions(opts ...ConfigurationOption) {
 	d.dockerConfigurationOptions = append(d.dockerConfigurationOptions, opts...)
+}
+
+// GetContainerConfig returns the container configuration and host configuration
+// used by the driver during container exec
+func (d *Driver) GetContainerConfig() (*container.Config, *container.HostConfig) {
+	return d.containerCfg, d.containerHostCfg
 }
 
 // Config returns the Docker driver configuration options
@@ -155,7 +163,7 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
 		env = append(env, fmt.Sprintf("%s=%v", k, v))
 	}
 
-	cfg := &container.Config{
+	d.containerCfg = &container.Config{
 		Image:        op.Image.Image,
 		Env:          env,
 		Entrypoint:   strslice.StrSlice{"/cnab/app/run"},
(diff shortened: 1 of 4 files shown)


driver/docker/docker.go
```diff
@@ -48,6 +50,12 @@ func (d *Driver) AddConfigurationOptions(opts ...ConfigurationOption) {
 	d.dockerConfigurationOptions = append(d.dockerConfigurationOptions, opts...)
 }
 
+// GetContainerConfig returns the container configuration and host configuration
+// used by the driver during container exec
+func (d *Driver) GetContainerConfig() (*container.Config, *container.HostConfig) {
```
silvin-lubecki, line 55: Almost nit-picking, but I'm a bit on the fence with this getter. I understand that it may be easier to use, but I think it's not a common golang pattern to return two unexposed fields at once 🤔 
WDYT @carolynvs ?

carolynvs: I see 3 different configuration variables that would be useful to query for unit tests: driver config (`d.config`), container config (`d.containerCfg`), and host config (`d.containerHostCfg`). Right now what is confusing is that this function is named like an accessor for just ContainerConfig, but it returns both container and host config. 

I recommend making 3 accessors, one for each, that returns copies instead of mutatable pointers to the config. Otherwise we open up another pattern for setting config on the driver that we didn't intend and more likely give people a way to write buggy tests. Splitting them up allows the tests to grab just the config variable they needed, without any extra processing.

vdice: Good call, thank you.  Split into two functions for accessing container config and container host config.  (If/when we desire a driver config accessor, we can add one in the future).   Also brought in [copystructure](https://godoc.org/github.com/mitchellh/copystructure) to assist in returning copies.  WDYT?

driver/docker/docker_test.go
```diff
@@ -38,14 +42,19 @@ func TestDriver_GetConfigurationOptions(t *testing.T) {
...
 			Privileged: true,
 		}
-		cfg, hostCfg := d.GetContainerConfig()
+
+		cfg, err := d.GetContainerConfig()
+		is.Nil(err)
 		is.Equal(expectedContainerCfg, cfg)
+
+		hostCfg, err := d.GetContainerHostConfig()
+		is.Nil(err)
```
carolynvs, line 57: This is feedback on when to use `Nil(error)` vs `NoError(error)`. The first will give you:

```
Expected nil, but got: &errors.errorString{s:"carolyn was here"}
```

The second gives you:

```
Received unexpected error:
carolyn was here
```

I'm not asking that you change anything, just sharing why in the future you may want to pick `NoError` instead for the improved failure message.

driver/docker/docker.go
```diff
@@ -50,10 +51,36 @@ func (d *Driver) AddConfigurationOptions(opts ...ConfigurationOption) {
...
-	return d.containerCfg, d.containerHostCfg
+func (d *Driver) GetContainerConfig() (container.Config, error) {
+	cpy, err := copystructure.Copy(*d.containerCfg)
+	if err != nil {
+		return container.Config{}, err
+	}
+
+	containerCfg, ok := cpy.(container.Config)
+	if !ok {
+		return container.Config{}, err
```
carolynvs, line 64: Here you are returning `err` but it will be `nil` but the returned config doesn't include the configuration data requested which is an error. Shouldn't this return an `errors.New` instead? Same for the method below.

driver/docker/docker_test.go
vdice: `is.TooValuableToIgnore(feeback) = true` 😄

driver/docker/docker.go
vdice: Indeed, thank you for catching.  Updated.

PR #204, opened 2020-04-16
ref(bundle.go): only validate/return value if param applies to provided action

vdice:
* adds functionality in the `ValuesOrDefaults` method to skip parameter validation and value lookup if it does not apply to the provided action
* in doing so, updates the function signature to now pass in the action

### DIFF
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -144,7 +144,7 @@ func TestValuesOrDefaults(t *testing.T) {
 		},
 	}
 
-	vod, err := ValuesOrDefaults(vals, b)
+	vod, err := ValuesOrDefaults(vals, b, "install")
 
 	is.NoError(err)
 	is.True(vod["enabled"].(bool))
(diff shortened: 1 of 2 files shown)


bundle/bundle_test.go
```diff
@@ -208,11 +208,65 @@ func TestValuesOrDefaults_Required(t *testing.T) {
...
+				Definition: "param-with-default",
+			},
+			"param-with-default-and-override": {
+				Definition: "param-with-default-and-override",
+			},
+		},
+	}
+
+	res, err := ValuesOrDefaults(vals, b, "install")
+	is.NoError(err)
```
carolynvs, line 263: suggestion: I recommend using a require here so that it doesn't continue trying to check the remaining asserts when an error occurs.

bundle/bundle_test.go
```diff
@@ -214,17 +214,18 @@ func TestValuesOrDefaults_Required(t *testing.T) {
...
 	}
+
 	b := &Bundle{
 		Definitions: map[string]*definition.Schema{
 			"param-with-default-not-applicable": {
 				Type:    "string",
 				Default: "foo",
 			},
-			"required-param-not-applicable": {
+			"required-param-with-default-not-applicable": {
```
carolynvs, line 228: Was this an intended change? I don't see a default set for this definition?

vdice: 🤦 Not intended; thank you! Should be fixed now.

PR #205, opened 2020-04-22
fix(docker.go): check nil pointer case; export ApplyConfigurationOptions

vdice:
* Avoids nil pointer panic if configuration uninitialized at time of retrieval
* Exports `ApplyConfigurationOptions` method, handy for consumers of this pkg

Q for reviewers: Should we explicitly error out if the configuration is uninitialized?  Or return an empty struct and no error, as is currently done in this PR?


vdice:
Closing this as I think I'll put up a replacement PR wherein the container config fields on the Driver object are no longer pointers... simplifies logic and is much more consumer-friendly.

PR #206, opened 2020-04-22
ref(docker.go): container config fields on Driver no longer pointers

vdice:
* update config fields on Driver struct to not be pointers

  *  When consuming the latest code for use in Porter (to add tests around Porter's configuration options in this area), I found it problematic that some struct fields were pointers (`containerCfg` and `containerHostCfg`).  I started to go down the path of creating methods in this library to return default Driver objects (`NewDriver()` would return a `Driver` with initialized/empty field references) but then I realized perhaps the code simplifications and ease of use that came with switching the fields to be actual values would be warranted here.  What do reviewers think?  Is there Golang best practice w/r/t fields in structs being pointers?  (Avoid?  Only use if initializer method(s) are provided?)
 
* export ApplyConfigurationOptions

  *  Porter will use this method to apply its configuration options without having to call `exec` on the driver in unit tests

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"archive/tar"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -20,6 +21,7 @@ import (
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/docker/pkg/stdcopy"
 	"github.com/docker/docker/registry"
+	"github.com/mitchellh/copystructure"
 )
 
 // Driver is capable of running Docker invocation images using Docker itself.
@@ -52,14 +54,34 @@ func (d *Driver) AddConfigurationOptions(opts ...ConfigurationOption) {
 
 // GetContainerConfig returns a copy of the container configuration
 // used by the driver during container exec
-func (d *Driver) GetContainerConfig() container.Config {
-	return d.containerCfg
+func (d *Driver) GetContainerConfig() (container.Config, error) {
+	cpy, err := copystructure.Copy(d.containerCfg)
+	if err != nil {
+		return container.Config{}, err
+	}
+
+	cfg, ok := cpy.(container.Config)
+	if !ok {
+		return container.Config{}, errors.New("unable to process container config")
+	}
+
+	return cfg, nil
 }
 
 // GetContainerHostConfig returns a copy of the container host configuration
 // used by the driver during container exec
-func (d *Driver) GetContainerHostConfig() container.HostConfig {
-	return d.containerHostCfg
+func (d *Driver) GetContainerHostConfig() (container.HostConfig, error) {
+	cpy, err := copystructure.Copy(d.containerHostCfg)
+	if err != nil {
+		return container.HostConfig{}, err
+	}
+
+	cfg, ok := cpy.(container.HostConfig)
+	if !ok {
+		return container.HostConfig{}, errors.New("unable to process container host config")
+	}
+
+	return cfg, nil
 }
 
 // Config returns the Docker driver configuration options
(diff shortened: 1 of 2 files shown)


driver/docker/docker.go
```diff
@@ -54,34 +52,14 @@ func (d *Driver) AddConfigurationOptions(opts ...ConfigurationOption) {
...
-	}
-
-	hostCfg, ok := cpy.(container.HostConfig)
-	if !ok {
-		return container.HostConfig{}, errors.New("unable to process container host config")
-	}
-
-	return hostCfg, nil
+func (d *Driver) GetContainerHostConfig() container.HostConfig {
+	return d.containerHostCfg
```
carolynvs, line 62: There is a difference between what you had before and what this does here.

Previously, your code made a deep copy of the entire config, ensuring that modifying the returned config could not affect the config used in `exec`. In this new function, it makes a copy but not how you may expect, and the calling code could use it to change the config, for example changing the data in the backing array behind the `CapAdd` slice.

```golang
// Assume that CapAdd already had SYS_PTRACE, the caller could alter it to be SYS_ADMIN instead without going through our defined interface of AddConfigurationOptions
hostCfg1 := d.GetContainerHostConfig()
hostCfg1.CapAdd[0] = "SYS_ADMIN"
hostCfg2 := d.GetContainerHostConfig()
is.Equal(hostCfg1, hostCfg2)
```

This is a contrived example, but is something that sometimes comes up when returning a slice or value that contains a slice, and then later we are wondering why the contents of the _original_ slice have changed inadvertently. It can be hard to track down, but doing a deep copy can avoid stuff like this. It's something that we should think about if we want to remove that protection or if we don't think it matters for this use case / library. 

Maybe we care, maybe we don't. 😀 I wanted to make sure you understood the difference in behavior between the two pieces of code. Otherwise previously if you had only cared about converting from a pointer to a value, there were cheaper methods so I had assumed you cared about solving this problem too.

vdice: Thank you.  I had the misunderstanding that by changing the field to a value, the deep copy was no longer applicable.  You've helped me see that a deep copy is applicable whenever the field struct itself contains slice fields, regardless of if we access by value or by pointer.  Therefore, I've updated to maintain the deep copy logic when accessing, as I think this is the prudent way to go, for reasons you've illustrated!

driver/docker/docker_test.go
```diff
@@ -31,15 +30,15 @@ func TestDriver_GetConfigurationOptions(t *testing.T) {
 	})
 
 	t.Run("configuration options", func(t *testing.T) {
-		d.containerCfg = &container.Config{}
-		d.containerHostCfg = &container.HostConfig{}
+		d := &Driver{}
```
carolynvs, line 33: 💯 this is much cleaner to work with

driver/docker/docker.go
```diff
@@ -293,9 +293,10 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
 	return opResult, err
 }
 
-func (d *Driver) applyConfigurationOptions() error {
+// ApplyConfigurationOptions applies the configuration options set on the driver
+func (d *Driver) ApplyConfigurationOptions() error {
```
silvin-lubecki, line 297: Question: I wonder what is the use case in porter to export this function? As dockerConfigurationOptions is not exported so noone can add new options, this function shouldn't be called twice? Or maybe from an API point of view this function should then have a variadic parameter with new options to apply?

vdice: The use case is to enable applying these configuration options outside of/without invoking the `exec` method where they otherwise will be applied.  Admittedly, this was originally motivated by unit testability, when we needed to check the config options that Porter adds without running exec (i.e. without actually invoking Docker).

However, I'd challenge the notion that this function shouldn't be called twice (or more). What would the fallout be if run multiple times?  The results of running `ApplyConfigurationOptions` should be idempotent unless the actual config option functions are updated (via the pre-existing `AddConfigurationOptions` method).  Speaking of which, you are right that `dockerConfigurationOptions` is not exported but API consumers can indeed add/modify options, and have been able to for some time, via the aforementioned [AddConfigurationOptions](https://github.com/cnabio/cnab-go/blob/master/driver/docker/docker.go#L50-L53)

silvin-lubecki: You are right I forgot this function! Ok to me then :+1:

PR #207, opened 2020-05-04
Fix ULID generation

carolynvs:
* The entropy generator needs to be reused between ULIDs to ensure that each ULID is monotonically increasing.
* We should use the monotonic function so that ULIDs generated during the same millisecond can still be sorted.
* Use a mutex in ULID since math.Rand is not thread-safe.

I was seeing in unit tests where I was generating ULIDs back-to-back that they were not sorting correctly. After digging through the issue queue, it turns out the example on the README of the ulid library we use is not appropriate to copy and use. I believe this is what they recommend based on their replies to a few issues.

### DIFF
--- a/claim/claim.go
+++ b/claim/claim.go
@@ -88,10 +88,15 @@ func New(name string) (*Claim, error) {
 	}
 
 	now := time.Now()
+	revision, err := NewULID()
+	if err != nil {
+		return nil, err
+	}
+
 	return &Claim{
 		SchemaVersion: schemaVersion,
 		Installation:  name,
-		Revision:      ULID(),
+		Revision:      revision,
 		Created:       now,
 		Modified:      now,
 		Result: Result{
@@ -111,7 +116,7 @@ func (c *Claim) Update(action, status string) {
 	c.Result.Action = action
 	c.Result.Status = status
 	c.Modified = time.Now()
-	c.Revision = ULID()
+	c.Revision = MustNewULID()
 }
 
 // Result tracks the result of an operation on a CNAB installation
@@ -154,9 +159,23 @@ var ulidMutex sync.Mutex
 // that each ULID is monotonically increasing.
 var ulidEntropy = ulid.Monotonic(rand.New(rand.NewSource(time.Now().UnixNano())), 0)
 
-// ULID generates a string representation of a ULID.
-func ULID() string {
+// MustNewULID generates a string representation of a ULID and panics on failure
+// instead of returning an error.
+func MustNewULID() string {
+	result, err := NewULID()
+	if err != nil {
+		panic(err)
+	}
+	return result
+}
+
+// NewULID generates a string representation of a ULID.
+func NewULID() (string, error) {
 	ulidMutex.Lock()
 	defer ulidMutex.Unlock()
-	return ulid.MustNew(ulid.Timestamp(time.Now()), ulidEntropy).String()
+	result, err := ulid.New(ulid.Timestamp(time.Now()), ulidEntropy)
+	if err != nil {
+		return "", errors.Wrap(err, "could not generate a new ULID")
+	}
+	return result.String(), nil
 }
--- a/claim/claim.go
+++ b/claim/claim.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/rand"
 	"regexp"
+	"sync"
 	"time"
 
 	"github.com/oklog/ulid"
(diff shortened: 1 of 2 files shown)


claim/claim_test.go
```diff
@@ -209,3 +210,24 @@ func TestClaimSchema(t *testing.T) {
...
+	for i := 0; i < 100; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			var last string
+			for j := 0; j < 1000; j++ {
+				next := ULID()
+				if strings.Compare(next, last) != 1 {
+					t.Fatal("generated a ULID that was not monotonically increasing")
+				}
```
vdice, line 228: We need to set `last` to `next` here or somewhere before we exit this for loop, right?  As I'm reading this, currently it seems `last` will always be an empty string.

carolynvs: bwahahahhahaha 🤦 Thanks Vaughn!

claim/claim.go
```diff
@@ -151,3 +145,18 @@ func (c Claim) Validate() error {
...
+
+// ulidEntropy must be set once and reused when generating ULIDs, to guarantee
+// that each ULID is monotonically increasing.
+var ulidEntropy = ulid.Monotonic(rand.New(rand.NewSource(time.Now().UnixNano())), 0)
+
+// ULID generates a string representation of a ULID.
+func ULID() string {
+	ulidMutex.Lock()
+	defer ulidMutex.Unlock()
+	return ulid.MustNew(ulid.Timestamp(time.Now()), ulidEntropy).String()
```
jlegrone, line 161: I didn't realize `ULID` could panic before now; obviously this is a separate discussion but it might be worth a breaking change to this package's API to follow the "Must_" naming convention. Perhaps `NewULID() (string, error)` and `MustNewULID() string` would be most idiomatic?

carolynvs: Thanks for calling out the panic. I liked your suggestion, and made those two functions and switched any code that could use NewULID over to it. 

I'm redoing our claims code right now and once I'm done it should all use NewULID. Changing that here would have been too much though since claim.Update doesn't return an error.

claim/claim.go
```diff
@@ -151,3 +150,32 @@ func (c Claim) Validate() error {
...
+	result, err := NewULID()
+	if err != nil {
+		panic(err)
+	}
+	return result
+}
+
+// NewULID generates a string representation of a ULID.
+func NewULID() (string, error) {
+	ulidMutex.Lock()
```
silvin-lubecki, line 174: I'm a bit concerned with locking everything everytime we update or create a new Claim.
The ULID documentation says the following:

> Please note that rand.Rand from the math package is not safe for concurrent use. Instantiate one per long living go-routine or use a sync.Pool if you want to avoid the potential contention of a locked rand.Source as its been frequently observed in the package level functions.

What about creating a new instance of rand and entropy each time we need it instead?
It may not make sense from a cryptographic point of view, but I think it's not the point here (otherwise we would use cryptographic rand instead?)
What about the following ?
```go
func NewULID() (string, error) {
    	now := time.Now()
	entropy := ulid.Monotonic(rand.New(rand.NewSource(now.UnixNano())), 0)
        result, err := ulid.New(ulid.Timestamp(now), entropy)
  	if err != nil {
		return "", errors.Wrap(err, "could not generate a new ULID")
	}
        return result.String(), nil
}
```

carolynvs: That doesn't yield increasing ulids unfortunately when called successively within small timeframes. You have to keep the `entropy` variable and reuse it between calls to `ulid.New` in order to have that work. Without this, unit tests that generate ulids back to back fail to sort properly. In a batch system (not a CLI with a user typing behind it), this could also be a problem.

I just took your implementation and ran it through `TestNewULID_ThreadSafe` which also asserts that each new ulid increases and it fails.

Below are some relevant issues from the ulid library. If you see a better recommended implementation, maybe you can submit a fix for what we have? Because what we are using is borked. 😀 

https://github.com/oklog/ulid/issues/8
https://github.com/oklog/ulid/issues/33

silvin-lubecki: > That doesn't yield increasing ulids unfortunately when called successively within small timeframes.

Indeed 🤦 

> Because what we are using is borked. 😀

Well I guess your implementation is then the less worst option 😄 But I think we then need make sure that the `NewULID` and `MustNewULID` function documentations stated explicitly that they are thread safe and use a global mutex.

carolynvs: Haha, yeah that's how I felt prototyping out solutions. "This is the least worst" 😀 

I'll update the documentation to include that explanation.

radu-matei:
Oh wow.. 
So to recap, generating an ULID can panic, and the operation is not thread safe.

Thanks **a lot** for catching this, I think we can create a new release for this alone once we merge.

PR #208, opened 2020-05-18
Fix linter settings and errors

radu-matei:
The `goimports` settings in the linter options were wrong, meaning `goimports` was not running properly.

This PR fixes that, as well as the linter errors.

### DIFF
--- a/golangci.yml
+++ b/golangci.yml
@@ -4,16 +4,16 @@ run:
 linters:
   disable-all: true
   enable:
-  - gofmt
-  - goimports
-  - golint
-  - gosimple
-  - ineffassign
-  - misspell
-  - unused
-  - deadcode
-  - govet
+    - gofmt
+    - goimports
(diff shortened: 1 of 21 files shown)


golangci.yml
```diff
@@ -4,16 +4,16 @@ run:
...
+    - ineffassign
+    - misspell
+    - unused
+    - deadcode
+    - govet
 
 linters-settings:
   goimports:
-    local-prefixes: github.com/radu-matei/cnab-go
+    local-prefixes: github.com/cnabio/cnab-go
```
carolynvs, line 19: 😂

vdice: throwback!

vdice:
I'm confused as to why the az pipeline is failing on this change 🤔

radu-matei:
Rebased, but for some reason the build is not triggering..

PR #209, opened 2020-05-19
Use GoTool Task

carolynvs:
Azure Devops moved where new versions of go (1.11+) are installed and no longer recommends using the scripts we had for setting up go. They have a Task that they want you to use to pick the exact version of go you want and then use it to go get your code.

Our builds were failing because as part of this GOROOT moved from `/usr/local/go1.13` to `/usr/local/go1.13.PATCH` which since we don't actually specify patch, was a bit hard to intuit. 😅

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/go?view=azure-devops&tabs=go-current#set-up-go

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,22 +1,10 @@
 pool:
   vmImage: 'Ubuntu 16.04'
 
-variables:
-  GOBIN:  '$(GOPATH)/bin' # Go binaries path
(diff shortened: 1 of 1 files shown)


carolynvs:
I have simplified this further to only use the GoTools task, and not use the Go task like their doc suggests. A MASTERPIECE! ✨

Issue #210, opened 2020-05-28
How to determine which outputs to retrieve?


carolynvs:
In the updated claims spec, we don't necessarily retrieve all the outputs anymore and store them in the claim. It's up to the tool to say which outputs it wants to persist. Additionally the claim result stores a map of the generated output names to an object that contains metadata about the outputs, currently we just have one piece of info we want: contentDigest.

I'm just thinking about the scenario where the tool does NOT want all the outputs persisted and if we should also avoid retrieving those outputs from the container too? If the tool doesn't want the output persisted, does that mean we don't pull it, generate the content digest and store that on the claim result?

Tagging in @jlegrone since this scenario is his use case.

PR #212, opened 2020-05-29
Rewrite Claims to support updated spec

carolynvs:
This is basically a rewrite, sorry. It updates our claims implementation to support the changes to the claims spec per [cnab-claim-1.0.0-DRAFT+b5ed2f3](https://github.com/cnabio/cnab-spec/releases/tag/cnab-claim-1.0.0-DRAFT%2Bb5ed2f3)

I apologize that this is a monolith PR. I found it impossible to break up due to the nature of the changes to the spec and how it impacted the codebase.

This is not an exhaustive list of changes, see the spec for that but here are the major changes:
* Split [Result](http://cnab.io/schema/cnab-claim-1.0.0-DRAFT+b5ed2f3/claim-result.schema.json) from [Claim](http://cnab.io/schema/cnab-claim-1.0.0-DRAFT+b5ed2f3/claim.schema.json)
* Do not persist outputs on Claim. Now they are persisted independently, and not all outputs have to be persisted, it's up to the tooling.
* Store output's content digest on the Result (see #210 for an open question about if we need to store every output's digest or just the ones we are persisting).
* Claim represents all the inputs to an operation, it doesn't change after an operation is performed. Result contains the status of the operation, and there can now be more than one so that we can keep track of its progress, e.g. Pending -> Running -> Succeeded.
* We don't reuse the same claim document for persistence anymore, preserving an audit trail / history. `crud.Store` and its implementations had to change it's API to support this, and querying by a "tag" or "foreign key" so that we could look for records by its key (e.g. claim id) or find all records by the "foreign key" such as installation name. While also supporting data types like credentials that do not need this. 
  The solution caters to the lowest common denominator, filesystem, and in the future we could introduce interfaces to help declare that an implementation supports querying (basically any other storage type like mongo or a real database).
* Expanded the claim store to support Result and Outputs and a new type and isn't persisted but is useful for everyone: installation. Installation represents an installation of a bundle (a set of claims).
* Claim store has built-in support for encrypting claim data and sensitive outputs. Claim should have always been encrypted or securely stored per the spec because they store sensitive data, such as database passwords. Now it's easier/encouraged for tools to implement that.
* Collapsed the implementation for actions into a single class `Action` without an interface since it wasn't being used. Action now has support for working with claims to help coordinate properly persisting them, especially when errors occur.
* Added example godoc demonstrating how to use Action and persist claim data properly in various scenarios.
* The logic around managing and defaulting outputs was simplified. Drivers don't have to deal with defaulting missing outputs or checking or missing required outputs, the runtime handles that now. Outputs are managed by name on OperationResult.

🚨 Based on PR #211, will need to rebase when that merges.


claim/claimstore_test.go
```diff
@@ -169,8 +226,103 @@ func TestClaimStore_HandlesNotFoundError(t *testing.T) {
...
+	err = s.SaveOutput(password)
+	require.NoError(t, err, "SaveOutput failed")
+
+	// Verify that password was encrypted at rest
+	encryptedOutputB, err := backingStore.Read(ItemTypeOutputs, s.outputKey(r.ID, password.Name))
+	require.NoError(t, err, "could not read raw output data")
+	decryptedOutputB, err := b64decode(encryptedOutputB)
+	require.NoError(t, err, "failed to decrypt raw output data")
+	assert.Equal(t, string(password.Value), string(decryptedOutputB), "decrypted output doesn't match the original output")
+
```
vdice, line 309: Do we want to also read directly from the helper (`s.ReadOutput(...)`) here, to confirm the `password` output is returned decrypted as expected?

claim/installation.go
```diff
@@ -0,0 +1,126 @@
...
+func (i Installation) GetLastOutputs(p Provider) (*Outputs, error) {
+	var results Results
+
+	claims, err := p.ReadAllClaims(i.Name)
+	if err != nil {
+		return nil, err
+	}
+	i.Claims = claims
+
+	// TODO: (carolynvs) retrieve data concurrently
```
vdice, line 85: Just a note to perhaps add to the list of areas under the general 'introduce go routines' follow-up as mentioned elsewhere.

claim/provider.go
```diff
@@ -0,0 +1,72 @@
...
+	// ReadInstallationStatus returns the specified Installation with the last Claim and its last Result loaded.
+	ReadInstallationStatus(installation string) (Installation, error)
+
+	// ReadAllInstallationStatus returns all Installations with the last Claim and its last Result loaded.
+	ReadAllInstallationStatus() ([]Installation, error)
+
+	// ReadClaim returns the specified Claim.
+	ReadClaim(claimID string) (Claim, error)
+
+	// ReadAllClaim returns all claims associated with an Installation, sorted in ascending order.
```
vdice, line 33:
```suggestion
	// ReadAllClaims returns all claims associated with an Installation, sorted in ascending order.
```

claim/provider.go
```diff
@@ -0,0 +1,72 @@
...
+	ReadAllClaims(installation string) ([]Claim, error)
+
+	// ReadLastClaim returns the last result associated with an Installation.
```
vdice, line 36:
```suggestion
	// ReadLastClaim returns the last claim associated with an Installation.
```

claim/provider.go
```diff
@@ -0,0 +1,72 @@
...
+	ReadOutput(claim Claim, result Result, outputName string) (Output, error)
+
+	// SaveClaim persists the specified claim.
+	// Associated results, Claim.Results, must be persisted separately with SaveResult.
+	SaveClaim(claim Claim) error
+
+	// SaveResult persists the specified result.
+	SaveResult(result Result) error
+
+	// SaveOutput persists the
```
vdice, line 58: Unfinished comment here...

claim/result.go
```diff
@@ -0,0 +1,112 @@
...
+	// Id of the result.
+	ID string `json:"id"`
+
+	// ClaimId associated with the claim that generated the result.
+	ClaimID string `json:"claimId"`
+
+	// Claim associated with the Result.
+	// This is not stored in the Result document but can be loaded onto
+	// the Result to build an in-memory hierarchy.
+	// TODO: (carolynvs) can we keep this? if so remove output.claim and use this everywhere instead
```
vdice, line 21: Does this still need to be addressed?

claim/result.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (r Results) Less(i, j int) bool {
+	return r[i].ID < r[j].ID
+}
+
+func (r Results) Swap(i, j int) {
+	r[i], r[j] = r[j], r[i]
+}
+
+// Outputs is the output metadata from an operation.
```
vdice, line 77:
```suggestion
// OutputMetadata is the output metadata from an operation.
```

claim/result.go
```diff
@@ -0,0 +1,112 @@
...
+		output = map[string]string{
+			OutputContentDigest: contentDigest,
+		}
+		o[outputName] = output
+		return nil
+	}
+
+	outputMetadata, ok := output.(map[string]string)
+	if !ok {
+		return errors.Errorf("cannot set the claim result's Outputs[%s][%s] because it is not type map[string]string but %T", outputName, OutputContentDigest, output)
```
vdice, line 107:
```suggestion
		return errors.Errorf("cannot set the claim result's OutputMetadata[%s][%s] because it is not type map[string]string but %T", outputName, OutputContentDigest, output)
```

driver/command/command_nix_test.go
```diff
@@ -114,7 +114,7 @@ func TestCommandDriverOutputs(t *testing.T) {
 			},
 		}
 		_, err := cmddriver.Run(&op)
-		assert.Errorf(t, err, "Command driver (test-outputs-missing.sh) failed for item: /cnab/app/outputs/output2 no output value found and no default value set")
+		assert.NoError(t, err)
 	}
 	CreateAndRunTestCommandDriver(t, name, content, testfunc)
 	// Test for an output missing with default value present
```
vdice, line 120: I wonder if we just update the comment here (maybe elsewhere) to clarify that now, even if a default for an output is declared in the bundle, it won't necessarily show up in the Operation Result's outputs list unless actually produced during the operation?

driver/debug/debug_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var _ driver.Driver = &Driver{}
+
+func TestDebugDriver_Handles(t *testing.T) {
+	d := &Driver{}
+	is := assert.New(t)
+	is.NotNil(d)
+	is.True(d.Handles(driver.ImageTypeDocker))
+	is.True(d.Handles("anything"))
```
vdice, line 20: 💪

driver/driver.go
```diff
@@ -50,8 +50,41 @@ type ResolvedCred struct {
...
+	// Logs is the combined logs from the bundle execution.
+	Logs bytes.Buffer
+
+	// Error is any errors from executing the operation.
+	Error error
+}
+
+// SetDefaultOutputValues for an output when it does not exist and it has a
+// non-empty default value.
+func (r *OperationResult) SetDefaultOutputValues(op Operation) error {
```
vdice, line 65: I may have missed a unit test for this function - is there one already?

action/action.go
```diff
@@ -17,15 +21,160 @@ import (
...
+	Claims         claim.Provider
+	Driver         driver.Driver
+	SaveAllOutputs bool
+	SaveOutputs    []string
+	SaveLogs       bool
+}
+
+// New creates an Action.
+// - Driver to execute the operation with.
+// - Claim Provider for persisting the claim data. If not set, calls to Save* will error.
```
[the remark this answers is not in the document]
carolynvs: I believe that DataDog uses their own optimized persistence layer and I didn't want to prevent them from using the action package anymore. Adding @jlegrone to confirm.

This is something that would be great to collaborate on in a follow-up and figure out where we need to have compromises like this and where we can improve the abstractions so that everyone is happy using them. But for the first go at it, I kept things open "minded" to avoid problems with known implementations.

action/action.go
```diff
@@ -17,15 +21,160 @@ import (
...
+}
+
+// saveClaimWithStatus saves a claim and a result with the specified status.
+func (a Action) saveClaimWithStatus(c claim.Claim, status string) error {
+	r, err := c.NewResult(status)
+	if err != nil {
+		return err
+	}
+
+	err = r.Validate()
```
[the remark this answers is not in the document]
carolynvs: I thought about that for bit but I would recommend finding a way to eliminate the error from the return parameters instead of doubling down on it.

I think we need to find a solution for creating ULIDs that doesn't have errors as a possibility. We rely on these documents to store critical data and logs, and not knowing if we even can store them in some cases is a problem.

I have ugly hacks in my back pocket to work around this*, but really we shouldn't be having to workaround it in the first place 😝  and should be able to safely generate a ULID and persist data about the state of the system.

\* essentially generate the ULIDs we need before we attempt the operation and munge the signature of these constructors to take the ULID up-front

claim/claim.go
```diff
@@ -75,68 +109,100 @@ func GetDefaultSchemaVersion() (schema.Version, error) {
...
 
-// New creates a new Claim initialized for an installation operation.
-func New(name string) (*Claim, error) {
-
-	if !ValidName.MatchString(name) {
-		return nil, fmt.Errorf("invalid installation name %q. Names must be [a-zA-Z0-9-_]+", name)
+// New creates a new Claim initialized for an operation.
+func New(installation string, action string, bun bundle.Bundle, parameters map[string]interface{}) (Claim, error) {
+	if !ValidName.MatchString(installation) {
+		return Claim{}, fmt.Errorf("invalid installation name %q. Names must be [a-zA-Z0-9-_]+", installation)
```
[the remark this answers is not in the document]
carolynvs: I didn't write/change that code as part of this PR. Let's do that in a follow-up?

claim/claimstore.go
```diff
@@ -2,82 +2,427 @@ package claim
...
+	names, err := s.ListInstallations()
+	if err != nil {
+		return nil, err
+	}
+
+	installations := make([]Installation, 0, len(names))
+	for _, name := range names {
+		installation, err := s.ReadInstallationStatus(name)
+		if err != nil {
+			// TODO: (carolynvs) for any of these ranges, return some results instead of nothing when one fails
```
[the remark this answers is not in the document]
carolynvs: the `// TODO: (carolynvs)` comments are things that I need to either create follow-up issues for or address in this PR.

I think I have whittled down all the comments at this point to large problems that were unsolved in the original code, so I'll be turning the remaining handful into follow-ups. For example this problem here is one we had in the original code, how do we return some data when iterating and we hit an error while communicating that one or more of the items was not retrieved?

claim/claimstore.go
```diff
@@ -2,82 +2,427 @@ package claim
...
 	return claim, err
 }
 
-// ReadAll retrieves all of the claims.
-func (s Store) ReadAll() ([]Claim, error) {
-	results, err := s.backingStore.ReadAll(ItemType)
+func (s Store) ReadAllClaims(installation string) ([]Claim, error) {
+	items, err := s.backingStore.ReadAll(ItemTypeClaims, installation)
 	if err != nil {
+		// TODO: handle installation not found
```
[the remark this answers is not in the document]
carolynvs: Oops, my TODO naming convention fell apart. 🤦‍♀️ These should have all been dealt with by now I'll do a pass for these nameless todos and figure them out.

claim/claimstore.go
```diff
@@ -2,82 +2,427 @@ package claim
...
+	sensitive, err := output.Claim.Bundle.IsOutputSensitive(output.Name)
+	if err != nil {
+		return errors.Wrapf(err, "could not determine if the output %q is sensitive", output.Name)
+	}
+
+	data := output.Value
+	if sensitive {
+		data, err = s.encrypt(output.Value)
+		if err != nil {
+			// TODO (carolynvs) make all of the error messages provide context like this one
```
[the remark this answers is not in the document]
carolynvs: I wanted to review all the error messages coming out of claim store and make sure if we had useful context info (like the name of the installation) that we were giving it in the error instead of giving vague errors. I'll do that and remove the TODO.

claim/claimstore.go
```diff
@@ -2,82 +2,427 @@ package claim
...
+	return s.backingStore.Save(ItemTypeOutputs, output.Result.ID, s.outputKey(output.Result.ID, output.Name), data)
+}
+
+func (s Store) DeleteInstallation(installation string) error {
+	claimIds, err := s.ListClaims(installation)
+	if err != nil {
+		return err
+	}
+
+	// Todo: go routines
```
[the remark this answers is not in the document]
carolynvs: Yes, I plan on making a follow-up issue for us to retrieve data concurrently.

claim/provider.go
carolynvs: https://www.youtube.com/watch?v=ZlIz0q8aWpA

claim/claimstore.go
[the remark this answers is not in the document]
carolynvs: I've created #215 to track this

claim/claimstore.go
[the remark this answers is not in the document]
carolynvs: I've created #214 to track this

action/action.go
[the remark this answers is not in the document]
carolynvs: I have created #216 to track

claim/claim.go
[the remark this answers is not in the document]
carolynvs: I have created #217 to track

claim/claim.go
```diff
@@ -147,8 +213,66 @@ func (c Claim) Validate() error {
...
+	if c.Installation == "" {
+		return errors.New("the installation must be set")
+	}
+
+	if c.Action == "" {
+		return errors.New("the action must be set")
+	}
+
+	// Check the action is built-in or defined as a custom action
+	if _, isBuildInAction := builtinActions[c.Action]; !isBuildInAction {
```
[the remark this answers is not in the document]
carolynvs: Ok fixed!

claim/claim_test.go
```diff
@@ -18,35 +19,104 @@ import (
 
 func TestNew(t *testing.T) {
 	// Make sure that the default Result has status and action set.
```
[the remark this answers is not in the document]
carolynvs: Fixed!

claim/claimstore.go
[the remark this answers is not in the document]
carolynvs: All TODO's have been addressed or turned into follow-up issues due to size/effort.

claim/claimstore.go
```diff
@@ -2,82 +2,427 @@ package claim
...
+	sort.Strings(claimIds)
+	lastClaimID := claimIds[len(claimIds)-1]
+
+	return s.ReadClaim(lastClaimID)
+}
+
+func (s Store) ReadResult(resultID string) (Result, error) {
+	bytes, err := s.backingStore.Read(ItemTypeResults, resultID)
+	if err != nil {
+		// TODO: handle installation/ claim / result not found
```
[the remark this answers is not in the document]
carolynvs: Fixed and tests added!

claim/claimstore.go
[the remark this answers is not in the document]
carolynvs: Fixed!

driver/command/command_nix_test.go
carolynvs: It doesn't show up in the intermediate OperationResult returned by the driver but it is set by the runtime ultimately before the caller gets the OperationResult back from the Action.

So I'm not sure where/how to document that or if that's necessary? Ultimately I moved that burden of defaulting unproduced outputs off the drivers and onto the runtime. I think that may make more sense to anyone coming in fresh (and who didn't see the original patch we had in place to get output defaults working). Drivers really care about executing the requested operation and less about enforcing every single side requirement of the runtime spec, which output defaults fall into (and probably shouldn't have ever been a responsibility of each driver in the first place).

But if you see a spot to call out where it's set, we can add something. I think cnab-go could really benefit from a higher level doc that just explains the flow of what is going on. I started something like that with the go doc examples, but we could use more that just explain how we have implemented the runtime perhaps?

driver/driver.go
carolynvs: I thought I had a couple but I can't find them anymore! 😱  I had them at the Run level to make sure they were being called (in TestAction_RunAction). I'll hunt them down from my git ref log and resurrect them.

claim/claimstore_test.go
carolynvs: Fixed!

claim/provider.go
carolynvs: Fixed!

driver/driver.go
carolynvs: Found and pushed!

carolynvs:
Whooooooo! I unskipped the test AND IT PASSED! The ultimate validation. 😀 Thanks for remembering that is was skipped.

carolynvs:
Rebased with changes from #211 that just merged.

action/example_install_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/cnabio/cnab-go/bundle"
+	"github.com/cnabio/cnab-go/claim"
+	"github.com/cnabio/cnab-go/credentials"
+	"github.com/cnabio/cnab-go/driver"
+	"github.com/cnabio/cnab-go/driver/lookup"
+	"github.com/cnabio/cnab-go/utils/crud"
+)
+
+// Install the bundle and only record the success/failure of the operation.
+func Example_install() {
```
radu-matei, line 16: nit: I think we've used `Example_Install` as convention for tests before (looking at, for example, `TestClaim_Validate`).

action/wellknown.go
```diff
@@ -1,55 +1,9 @@
 package action
 
-import (
-	"github.com/cnabio/cnab-go/claim"
-	"github.com/cnabio/cnab-go/credentials"
-	"github.com/cnabio/cnab-go/driver"
-)
-
 // Well known constants define the Well Known CNAB actions to be taken
 const (
```
radu-matei, line 4: Will the well known action list expand over time?
Does it make sense to keep this in a separate file?

claim/provider.go
```diff
@@ -0,0 +1,81 @@
...
+	// ListClaims returns Claim IDs associated with an Installation sorted in ascending order.
+	ListClaims(installation string) ([]string, error)
+
+	// ListResults returns Result IDs associated with a Claim, sorted in ascending order.
+	ListResults(claimID string) ([]string, error)
+
+	// ListOutputs returns the names of outputs associated with a result that
+	// have been persisted. It is possible for results to have outputs that were
+	// generated but not persisted see
+	// https://github.com/cnabio/cnab-spec/blob/master/400-claims.md#outputs for
```
radu-matei, line 17: Do we want this to be a permalink?

action/example_install_test.go
carolynvs: Oh yeah so I really wanted to name it that too, but then it doesn't show up in the godoc! 😅 It only shows up when lower-cased. I'm not sure why but that's why I named the whole file examples that way.

In the "doc" for godoc, they tell you to look at the sort package and that's what they do there too

https://blog.golang.org/examples
https://golang.org/src/sort/example_keys_test.go

If you know how to get it working with upper-case, let me know!

action/wellknown.go
carolynvs: Ah, probably not. I think we are done identifying well-known actions in the spec. Now that this file is just this const, I can move the definitions to action.go.

action/example_install_test.go
jlegrone: Have we tried `ExampleInstall`?

driver/debug/debug.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Driver prints the information passed to a driver
+//
+// It does not ever run the image.
+type Driver struct {
+	config map[string]string
+}
+
+// Run executes the operation on the Debug driver
+func (d *Driver) Run(op *driver.Operation) (driver.OperationResult, error) {
```
jlegrone, line 18: Can we pass a context object as well to handle things like deadlines/cancellation/tracing?

```suggestion
func (d *Driver) Run(ctx context.Context, op *driver.Operation) (driver.OperationResult, error) {
```

action/example_install_test.go
carolynvs: Yup! godoc doesn't like that either 🤷‍♀️

driver/debug/debug.go
carolynvs: I started making this change here, but I think it would be better in a second PR so we can really look over the changes to implementing the context handling in the drivers in isolation. I'll merge this then immediately open up the PR with my changes and tests so far and we can see what else needs to be done.

PR #213, opened 2020-05-29
Stop on nonzero exit code

carolynvs:
When you embed bash in yaml, make sure to stop on nonzero exit codes. Otherwise builds pass when they should have failed...

Signed-off-by: Carolyn Van Slyck <carolyn.vanslyck@microsoft.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,13 +49,13 @@ ifndef HAS_GOCOV_XML
 	go get github.com/AlekSi/gocov-xml
 endif
 ifndef HAS_GOCOV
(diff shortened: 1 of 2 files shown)


carolynvs:
Alrighty! I've managed to get the build showing red when it's really broken (which has been for a while) and then fix it... 

go mods shenanigans were involved. Don't ask why `go mods download` was necessary and just building directly pulled down the wrong package. I'm guessing it has something to do with the tides and packr force pushing a tag? 🤷‍♀️

PR #218, opened 2020-06-02
Fail build take two

carolynvs:
When I fixed adding go/bin to PATH, I accidentally moved instead of copied the lines to fail the script. Adding it back. 🤦‍♀️ 

So this does a check that I can yet again fail the build, and then fixes the installation of gocov, which isn't working on master, so we aren't publishing coverage results.


carolynvs:
Cool, we have test and code coverage again! 🥳 
Screen Shot 2020-06-02 at 11 11 40 AM

PR #219, opened 2020-06-04
ref(*): extract set, source, strategy and place into valuesource

vdice:
* Refactor to generalize the `Set`, `Source` and `Strategy` structs so that they can be used for other sets needing source value resolution, such as parameter sets.

Ref https://github.com/deislabs/porter/pull/1068


carolynvs:
Sorry just realized this is still in draft. Are there more changes to be made?

valuesource/strategy.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Set is an actual set of resolved values.
+// This is the output of resolving a parameter or credential set file.
+type Set map[string]string
+
+// ExpandCredentials expands the set into env vars and paths per the spec in the bundle.
+//
+// This matches the credentials required by the bundle to the credentials present
+// in the Set, and then expands them per the definition in the Bundle.
+func (s Set) ExpandCredentials(b *bundle.Bundle, stateless bool) (env, files map[string]string, err error) {
```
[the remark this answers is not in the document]
vdice: Good point.  I think it makes sense to make this change in this PR.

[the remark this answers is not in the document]
vdice: I moved the method into `action.go` and updated its signature (no longer a method on a `Set` struct; takes a `valuesource.Set` object as a param).  WDYT?

vdice:
Rebased off the latest changes.  Note I also added the following commit: https://github.com/cnabio/cnab-go/pull/219/commits/581ddfb876a10058004c3941d3757e096728fcce

This updates the invalid source verbiage to be more general, as the source value may not be for a credential (but could be for a parameter, as can be the case in Porter).

action/action.go
```diff
@@ -365,8 +365,8 @@ func getImageMap(b bundle.Bundle) ([]byte, error) {
 	return json.Marshal(imgs)
 }
 
-func opFromClaim(stateless bool, c claim.Claim, ii bundle.InvocationImage, creds credentials.Set) (*driver.Operation, error) {
-	env, files, err := creds.Expand(c.Bundle, stateless)
+func opFromClaim(stateless bool, c claim.Claim, ii bundle.InvocationImage, creds valuesource.Set) (*driver.Operation, error) {
+	env, files, err := expandCredentials(&c.Bundle, creds, stateless)
```
carolynvs, line 369: As a suggested simplification, now that claim.Bundle is a value, we can change the signature of `expandCredentials` to take a value too and not need to change it to a pointer here and elsewhere (like in the tests) anymore.

carolynvs:
I've been using this code in a test branch of Porter for a couple weeks now and it's been working well 👍

action/action.go
vdice: Good call; updated!

action/action_test.go
```diff
@@ -1051,3 +1050,84 @@ func TestSaveAction(t *testing.T) {
 		assert.Error(t, err, "the output should NOT be persisted")
 	})
 }
+
+func Test_expandCredentials(t *testing.T) {
```
radu-matei, line 1054: nit: is there a reason for the different naming of this test?
(i.e. why not `TestExpandCredentials`?)

valuesource/strategy.go
```diff
@@ -0,0 +1,108 @@
...
+	var raw map[string]string
+	err := json.Unmarshal(data, &raw)
+	if err != nil {
+		return err
+	}
+	return s.unmarshalRaw(raw)
+}
+
+func (s Source) MarshalYAML() (interface{}, error) {
+	// TODO: use https://github.com/ghodss/yaml so that we don't need json and yaml defined
```
radu-matei, line 97: Is there an open issue for this?

vdice: Thanks; created https://github.com/cnabio/cnab-go/issues/222 to track this item (not intro'd in this PR).

action/action_test.go
vdice: No particular reason for the test name so updated to `TestExpandCredentials` 👍

Issue #220, opened 2020-06-16
Handle context cancellation in the kubernetes driver


carolynvs:
A context object is now being passed to the drivers during `Run`. The kubernetes driver needs to take advantage of it to handle when the passed context times out or is cancelled.

jlegrone:
Do we need a similar issue for the docker driver?

PR #221, opened 2020-06-16
Pass context to drivers during bundle operation

carolynvs:
When we execute a bundle operation, pass a context so that they can handle cancellations and timeouts.

The kubernetes driver required enough changes that I created #220 to track updating it to actually use cxt.

TODO:
- [ ] create a context to pass into the action (in the examples)
- [ ] add tests for context cancellation being handled appropriately in action, i.e. we don't do anything when it's already been cancelled
- [ ] update action.Run signature

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -1,6 +1,7 @@
 package action
 
 import (
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
@@ -78,7 +79,7 @@ func (a Action) Run(c claim.Claim, creds credentials.Set, opCfgs ...OperationCon
 	}
 
 	var opErr *multierror.Error
-	opResult, err := a.Driver.Run(op)
+	opResult, err := a.Driver.Run(context.TODO(), op)
 	if err != nil {
 		opErr = multierror.Append(opErr, err)
 	}
--- a/action/action_test.go
+++ b/action/action_test.go
@@ -1,6 +1,7 @@
 package action
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"io/ioutil"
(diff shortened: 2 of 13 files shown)


action/action.go
```diff
@@ -78,7 +79,7 @@ func (a Action) Run(c claim.Claim, creds credentials.Set, opCfgs ...OperationCon
 	}
 
 	var opErr *multierror.Error
-	opResult, err := a.Driver.Run(op)
+	opResult, err := a.Driver.Run(context.TODO(), op)
```
carolynvs, line 82: I'm still working on how we want to introduce the context, leaning towards the first parameter to `Action.Run` unless there's a reason for opCfgs to be involved somehow?

action/action_test.go
```diff
@@ -32,7 +33,7 @@ type mockDriver struct {
 func (d *mockDriver) Handles(imageType string) bool {
 	return d.shouldHandle
 }
-func (d *mockDriver) Run(op *driver.Operation) (driver.OperationResult, error) {
+func (d *mockDriver) Run(context context.Context, op *driver.Operation) (driver.OperationResult, error) {
```
jlegrone, line 36: To avoid shadowing the package name

```suggestion
func (d *mockDriver) Run(ctx context.Context, op *driver.Operation) (driver.OperationResult, error) {
```

driver/command/command_nix_test.go
```diff
@@ -174,3 +178,40 @@ func TestCommandDriverOutputs(t *testing.T) {
...
+			Revision:    "01DDY0MT808KX0GGZ6SMXN4TW",
+			Environment: map[string]string{},
+			Files: map[string]string{
+				"/cnab/app/image-map.json": "{}",
+			},
+			Out:    &output,
+			Bundle: &bundle.Bundle{Name: "mybun"},
+		}
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
```
jlegrone, line 211: We probably want this instead?
```suggestion
		defer cancel()
```

action/action.go
jlegrone: Making context the first parameter is probably most idiomatic. From the [package docs](https://golang.org/pkg/context/):

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

carolynvs:
Closing because this is really out of date

PR #227, opened 2020-08-05
feat(docker): add verification of image digest(s)

vdice:
* Adds invocation image digest verification per the [CNAB Spec](https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#invocation-images) to the Docker driver.

The specific part of the spec that this PR currently implements is the following line:

`If a contentDigest field is present, a runtime MUST validate the image digest prior to executing an action.`

I didn't see how this would be done generically, e.g. outside of specific driver implementations, hence implementing it on the Docker driver only.  Am I right here?  If so, perhaps a follow-up could implement similar in the kubernetes driver.

Note some remaining TODOs/Qs inline.  Feel free to respond/answer in the form of code review comments.

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -3,12 +3,12 @@ package docker
 import (
 	"archive/tar"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	unix_path "path"
+	"strings"
 
 	"github.com/docker/cli/cli/command"
 	cliflags "github.com/docker/cli/cli/flags"
@@ -22,6 +22,9 @@ import (
 	"github.com/docker/docker/registry"
 	"github.com/mitchellh/copystructure"
 
+	"github.com/pkg/errors"
+
+	"github.com/cnabio/cnab-go/bundle"
 	"github.com/cnabio/cnab-go/driver"
 )
 
@@ -187,6 +190,17 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
 			return driver.OperationResult{}, err
 		}
 	}
+
+	ii, err := d.inspectImage(ctx, op.Image)
+	if err != nil {
+		return driver.OperationResult{}, err
+	}
+
+	err = d.validateImageDigest(op.Image, ii.RepoDigests)
+	if err != nil {
+		return driver.OperationResult{}, errors.Wrap(err, "image digest validation failed")
+	}
+
 	var env []string
 	for k, v := range op.Environment {
 		env = append(env, fmt.Sprintf("%s=%v", k, v))
@@ -206,16 +220,7 @@ func (d *Driver) exec(op *driver.Operation) (driver.OperationResult, error) {
 	}
 
 	resp, err := cli.Client().ContainerCreate(ctx, &d.containerCfg, &d.containerHostCfg, nil, "")
-	switch {
-	case client.IsErrNotFound(err):
-		fmt.Fprintf(cli.Err(), "Unable to find image '%s' locally\n", op.Image.Image)
-		if err := pullImage(ctx, cli, op.Image.Image); err != nil {
-			return driver.OperationResult{}, err
-		}
-		if resp, err = cli.Client().ContainerCreate(ctx, &d.containerCfg, &d.containerHostCfg, nil, ""); err != nil {
-			return driver.OperationResult{}, fmt.Errorf("cannot create container: %v", err)
-		}
-	case err != nil:
+	if err != nil {
 		return driver.OperationResult{}, fmt.Errorf("cannot create container: %v", err)
 	}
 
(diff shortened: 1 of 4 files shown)


driver/docker/docker.go
```diff
@@ -385,3 +390,52 @@ func generateTar(files map[string]string) (io.Reader, error) {
...
+	case err != nil:
+		return ii, errors.Wrapf(err, "cannot inspect image %s", image.Image)
+	}
+
+	return ii, nil
+}
+
+// validateImageDigest validates the operation image digest, if exists, against
+// the supplied repoDigests
+// TODO: or is it only acceptable for repoDigests to have one, exact match?
```
vdice, line 416: It sounds like, indeed, it is only acceptable for repoDigests to have one, exact match.  Will update.

vdice:
@carolynvs @radu-matei @silvin-lubecki if any of y'all have a spare moment, it would be great to get feedback on this PR, as it addresses an area where cnab-go isn't in adherence to the spec.

driver/docker/docker.go
```diff
@@ -385,3 +390,53 @@ func generateTar(files map[string]string) (io.Reader, error) {
...
+		switch count := len(repoDigests); {
+		case count == 0:
+			return fmt.Errorf("image %s has no repo digests", image.Image)
+		case count > 1:
+			return fmt.Errorf("image %s has more than one repo digest", image.Image)
+		}
+
+		repoDigest := repoDigests[0]
+		// RepoDigests are of the form 'imageName@sha256:<sha256>'; we parse out the digest itself for comparison
+		// TODO: is there a better (i.e. via docker api) way to extract the digest?
```
carolynvs, line 427: Here's something

```go
// RepoDigests are of the form 'imageName@sha256:<sha256>'; we parse out the digest itself for comparison
repoDigest := repoDigests[0]
ref, err := reference.ParseNormalizedNamed(repoDigest)
if err != nil {
	return err
}
digestRef, ok := ref.(reference.Digested)
if !ok {
	return fmt.Errorf("unable to parse repo digest %s", repoDigest)
}
digest := digestRef.Digest().String()
```

This is the updated unit test I used to verify

```go
func TestDriver_ValidateImageDigest(t *testing.T) {
	repoDigests := []string{
		"myreg/myimg@sha256:9d4178a3b6059e46a02e4c5b39ac26d51d91046faeb7bb9f3c3825f92adafc5a",
	}

	t.Run("no image digest", func(t *testing.T) {
		d := &Driver{}

		image := bundle.InvocationImage{}
		image.Image = "myreg/myimg"

		err := d.validateImageDigest(image, repoDigests)
		assert.NoError(t, err)
	})

	t.Run("image digest exists - no match exists", func(t *testing.T) {
		d := &Driver{}

		image := bundle.InvocationImage{}
		image.Image = "myreg/myimg"
		image.Digest = "sha256:8b92b7269f59e3ed824e811a1ff1ee64f0d44c0218efefada57a4bebc2d7ef6f"

		err := d.validateImageDigest(image, repoDigests)
		assert.NotNil(t, err, "expected an error")
		assert.Contains(t, err.Error(), "content digest mismatch")
	})

	t.Run("image digest exists - a match exists", func(t *testing.T) {
		d := &Driver{}

		image := bundle.InvocationImage{}
		image.Image = "myreg/myimg"
		image.Digest = "sha256:9d4178a3b6059e46a02e4c5b39ac26d51d91046faeb7bb9f3c3825f92adafc5a"

		err := d.validateImageDigest(image, repoDigests)
		assert.NoError(t, err)
	})
}
```

I had to switch to full digests because the docker library doesn't parse short digests.

driver/docker/docker.go
```diff
@@ -385,3 +390,53 @@ func generateTar(files map[string]string) (io.Reader, error) {
...
+		digest := parts[1]
+
+		if digest == image.Digest {
+			return nil
+		}
+		return fmt.Errorf("content digest mismatch: image %s has digest %s but the value should be %s according to the bundle file", image.Image, digest, image.Digest)
+	}
+	// TODO: if digest empty, do we want to provide a warning somehow? Spec says:
+	// If the contentDigest is not present, the runtime SHOULD report an error so the user is aware that there is no contentDigest provided.
+	return nil
```
carolynvs, line 441: I think this is something that the the tool should do. In developer loops, we fully expect the digest to be left off until it has stabilized. So they (tools like porter or duffle) are in a better position to bubble up and make that warning proper.

For example, allowing it to be empty when running from a local bundle (and only displaying a warning) but perhaps displaying a stronger error if there is no digest in a published bundle.

carolynvs (commented):
Looks good, I've provided some suggestions on how to use the docker library for the parsing.

driver/docker/docker.go
```diff
@@ -385,3 +390,53 @@ func generateTar(files map[string]string) (io.Reader, error) {
...
+		return ii, errors.Wrapf(err, "cannot inspect image %s", image.Image)
+	}
+
+	return ii, nil
+}
+
+// validateImageDigest validates the operation image digest, if exists, against
+// the supplied repoDigests
+func (d *Driver) validateImageDigest(image bundle.InvocationImage, repoDigests []string) error {
+	if image.Digest != "" {
```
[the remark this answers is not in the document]
vdice: 👍 I feel like I always miss this one; please keep reminding me!  Definitely like the clarity/readability improvement.

driver/docker/docker.go
vdice: Thank you, I've incorporated these changes.  Changing to full digests in the tests is also an improvement.

driver/docker/docker.go
vdice: I agree.  I've removed this TODO and we'll leave implementation up to tools.

driver/docker/docker.go
```diff
@@ -414,29 +413,31 @@ func (d *Driver) inspectImage(ctx context.Context, image bundle.InvocationImage)
...
 
-		if digest == image.Digest {
-			return nil
-		}
-		return fmt.Errorf("content digest mismatch: image %s has digest %s but the value should be %s according to the bundle file", image.Image, digest, image.Digest)
+	// RepoDigests are of the form 'imageName@sha256:<sha256>'; we parse out the digest itself for comparison
+	repoDigest := repoDigests[0]
+	ref, err := reference.ParseNormalizedNamed(repoDigest)
+	if err != nil {
+		return err
```
carolynvs, line 431: Sorry I gave you informal code. We should probably wrap this error and provide a better error message with context. Otherwise it won't be clear what reference is invalid.

carolynvs (commented):
LGTM just one nit to resolve if you don't mind

driver/docker/docker.go
vdice: Updated!  And deadbeef returns in a test for this scenario! 🎉

driver/docker/docker_test.go
```diff
@@ -113,15 +113,18 @@ func TestDriver_ValidateImageDigest(t *testing.T) {
...
 		d := &Driver{}
 
 		image := bundle.InvocationImage{}
 		image.Image = "myreg/myimg"
-		image.Digest = "sha256:d366a4665ab44f0648d7a00ae3fae139d55e32f9712c67accd604bb55df9d05a"
+		image.Digest = "sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321"
 
-		err := d.validateImageDigest(image, repoDigests)
-		assert.NoError(t, err)
+		badRepoDigests := []string{"myreg/myimg@sha256:deadbeef"}
```
carolynvs, line 123: THE RETURN OF DEADBEEFS 👍

carolynvs (changes requested):
LGTM for reals this time

carolynvs (approved):
github is not liking my approvals 

/force LGTM dagnabbit

radu-matei (approved):

PR #229, opened 2020-08-17
Add schemaVersion to credentialSet

carolynvs:
Add a schemaVersion field to credentialSet so that we can identify the spec version that the credential set adheres to. This was very useful for claims, we use it for data migrations when the claim spec changes, and it is best to have it _before_ we need to change the spec. Worst case we never change the spec again and we all win. 😄

The last change to the credential-set spec was https://github.com/cnabio/cnab-spec/releases/tag/cnab-credentialsets-1.0.0-DRAFT+b6c701f

### DIFF
--- a/credentials/credentialset.go
+++ b/credentials/credentialset.go
@@ -13,10 +13,17 @@ import (
 	"github.com/cnabio/cnab-go/valuesource"
 )
 
-// CNABSpecVersion represents the CNAB Spec version of the Credentials
-// that this library implements
-// This value is prefixed with e.g. `cnab-credentials-` so isn't itself valid semver.
-var CNABSpecVersion string = "cnab-credentialsets-1.0.0-DRAFT-b6c701f"
+const (
+	// DefaultSchemaVersion is the default SchemaVersion value
+	// set on new CredentialSet instances, and is the semver portion
+	// of CNABSpecVersion.
+	DefaultSchemaVersion = schema.Version("1.0.0-DRAFT+b6c701f")
+
+	// CNABSpecVersion represents the CNAB Spec version of the Credentials
+	// that this library implements
+	// This value is prefixed with e.g. `cnab-credentials-` so isn't itself valid semver.
+	CNABSpecVersion string = "cnab-credentialsets-" + string(DefaultSchemaVersion)
+)
 
 // CredentialSet represents a collection of credentials
 type CredentialSet struct {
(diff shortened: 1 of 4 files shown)


carolynvs:
So one thing I'm unsure about is that @vdice recently added support for parameter sets. It's not in the spec, but it would be great to see that added to it and cnab-go. If we did that, would the spec still be called credentialsets? They have different schema, the list of parameters goes under a field named `parameters` in a parameter set and in a credential set it has a field named `credentials`. Or would we have two specs? 🤷‍♀️

credentials/credentialset.go
```diff
@@ -8,12 +8,20 @@ import (
...
 )
 
+// CNABSpecVersion represents the CNAB Spec version of the Credentials
+// that this library implements
+// This value is prefixed with e.g. `cnab-credentials-` so isn't itself valid semver.
+var CNABSpecVersion string = "cnab-credentialsets-1.0.0-DRAFT-b6c701f"
+
 // CredentialSet represents a collection of credentials
 type CredentialSet struct {
+	// SchemaVersion is the version of the claim schema.
```
vdice, line 23: s/claim/credential (or credentialset)

vdice:
Good questions, @carolynvs .

I also noticed that an actual schema for a credentialset isn't yet present in the cnab-spec repo, beyond the in-line example/text in https://github.com/cnabio/cnab-spec/blob/master/802-credential-sets.md.  We can perhaps create an issue in the spec repo to create one.

I'm conflicted on whether or not to try to combine parameter set and credential set into one 'valuesource' schema or leave them separate.  The latter approach perhaps has benefits in comprehension and, theoretically, flexibility in the future, if, for example, a parameter set adds or removes sources that a credential set lacks or wishes to keep...

I'll create in issue in this repo around the general area of adding in parameter set definitions/logic and include the question/task around pinning to a correlating schema and/or spec tag.

carolynvs:
@vdice What do you think of this alternate pattern for getting at the DefaultSchemaVersion so that we don't have to check for parse errors at runtime?

https://github.com/cnabio/cnab-go/pull/229/files#diff-2c1fa05be1e11946f6192fea4d31108dR20

vdice:
@carolynvs Sorry for the late reply!  Somehow missed the last comment.  👍 I love the alternate approach you've gone with -- great idea!

carolynvs:
@radu-matei What do you think? Right now I'm writing Porter to work with 2 specs, since that is how CNAB is written today.

radu-matei (approved):

PR #231, opened 2020-08-19
fix(filesystem.go): also delete group directory

vdice:
* Adds logic to remove the parent (group) directory if the item being deleted is the last one in that dir

Previously, group directories -- where a group is a named (ULID or installation name) directory under the item type directories of `claims`, `results` or `outputs` -- would linger after deleting an installation.

### DIFF
--- a/utils/crud/filesystem_test.go
+++ b/utils/crud/filesystem_test.go
@@ -3,6 +3,7 @@ package crud
 import (
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


utils/crud/filesystem_test.go
```diff
@@ -12,24 +13,86 @@ import (
...
+			require.NoError(t, s.Delete(testItemType, keys[0]))
+
+			// Verify list count
+			list, err = s.List(testItemType, tc.group)
+			require.NoError(t, err, "expected no error when listing directly from the item type directory")
+			require.Len(t, list, len(keys)-1)
+
+			// Verify that the group/parent dir remains
+			groupDir, err := os.Stat(filepath.Join(tmdir, testItemType, tc.group))
+			require.NoError(t, err, "expected the group directory to exist")
```
carolynvs, line 66: suggestion: For the test case where group is empty, this error message is misleading

```suggestion
			require.NoError(t, err, "expected the group/parent directory to exist")
```

utils/crud/filesystem_test.go
```diff
@@ -12,24 +13,86 @@ import (
...
+			require.True(t, groupDir.IsDir())
+
+			// Delete last record
+			require.NoError(t, s.Delete(testItemType, keys[1]))
+
+			// Verify group/parent dir removed
+			if tc.group != "" {
+				_, err := os.Stat(filepath.Join(tmdir, testItemType, tc.group))
+				require.Contains(t, err.Error(), "no such file or directory",
```
carolynvs, line 75: Here's a way to test the error without relying on the error message:

```suggestion
				require.True(t, os.IsNotExist(err),
```

utils/crud/filesystem_test.go
```diff
@@ -12,24 +13,86 @@ import (
...
+					"expected the parent group directory to be removed")
+			}
+
+			// Verify group/parent dir removed and error received
+			// or list is empty
+			list, err = s.List(testItemType, tc.group)
+			if tc.group != "" {
+				require.EqualError(t, err, "File does not exist",
```
carolynvs, line 83: Here's how you can avoid hard checking the error message:

```suggestion
				require.Equal(t, ErrRecordDoesNotExist, err,
```

vdice:
@radu-matei do you think you'd have a spare moment to check this PR out?

radu-matei (approved):

PR #232, opened 2020-10-01
Fix invocation image digest validation

carolynvs:
Bundles pushed with cnab-to-oci will have invocation images with more than one repoDigest, which is valid. One is for the original invocation image that was pushed. The other is for the relocated invocation image that is now inside the bundle repository.

https://porter.sh/distribute-bundles/#image-references-after-publishing

We should check all of the repoDigests and if a match is found, then it is valid.

This corrects a bug introduced in #227 that is present in [v0.14.0](https://github.com/cnabio/cnab-go/releases/tag/v0.14.0) where executing a bundle fail because bundles pushed with cnab-to-to-oci always have more than one repoDigest.

### DIFF
--- a/driver/docker/docker.go
+++ b/driver/docker/docker.go
@@ -417,27 +417,31 @@ func (d *Driver) validateImageDigest(image bundle.InvocationImage, repoDigests [
 		return nil
 	}
 
-	switch count := len(repoDigests); {
-	case count == 0:
+	if len(repoDigests) == 0 {
 		return fmt.Errorf("image %s has no repo digests", image.Image)
-	case count > 1:
-		return fmt.Errorf("image %s has more than one repo digest", image.Image)
 	}
 
-	// RepoDigests are of the form 'imageName@sha256:<sha256>'; we parse out the digest itself for comparison
-	repoDigest := repoDigests[0]
-	ref, err := reference.ParseNormalizedNamed(repoDigest)
-	if err != nil {
-		return fmt.Errorf("unable to parse repo digest %s", repoDigest)
-	}
-	digestRef, ok := ref.(reference.Digested)
-	if !ok {
-		return fmt.Errorf("unable to parse repo digest %s", repoDigest)
-	}
-	digest := digestRef.Digest().String()
+	for _, repoDigest := range repoDigests {
+		// RepoDigests are of the form 'imageName@sha256:<sha256>' or imageName:<tag>
+		// We only care about the ones in digest form
+		ref, err := reference.ParseNormalizedNamed(repoDigest)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


vdice (commented):
Thank you for the fix.  Ready to LGTM, just needs a few updates to some assertions in the docker_integration_test.go.

carolynvs:
Doh! Thanks, that's what I get for tweaking the error message at the last minute. 😊

carolynvs:
All fixed

vdice (approved):
LGTM!

driver/docker/docker.go
```diff
@@ -417,27 +417,31 @@ func (d *Driver) validateImageDigest(image bundle.InvocationImage, repoDigests [
...
-	if !ok {
-		return fmt.Errorf("unable to parse repo digest %s", repoDigest)
-	}
-	digest := digestRef.Digest().String()
+	for _, repoDigest := range repoDigests {
+		// RepoDigests are of the form 'imageName@sha256:<sha256>' or imageName:<tag>
+		// We only care about the ones in digest form
+		ref, err := reference.ParseNormalizedNamed(repoDigest)
+		if err != nil {
+			return fmt.Errorf("unable to parse repo digest %s", repoDigest)
```
silvin-lubecki, line 429: question: Should we error out here, or just ignore this case? 🤔 
I don't know if in this case we can have both digested reference AND tagged reference, but if we find the first one and just error out instead of trying the second one which is the one we want, we may return a validation error even if it's valid?

carolynvs: Did you intend to comment on line 434 instead of where we check if the reference is in digest form and if not continue? This error check on line 429 is for when the repository digest is an invalid reference entirely. Or at least that is my intent.

Assuming you meant line 434, in practice there will always be a tagged and digested reference for the invocation images when Porter uses cnab-to-oci. When Porter first creates the invocation image, it pushes it to a registry, `myuser/mybundle-installer:v1.0`, this allows us to get a repository digest for the image. Then it is copied to its final location at `myuser/mybundle@sha256:invocation-image-repo-digest`. At runtime when we pull the invocation image and inspect it we get two repository digests back:

* `myuser/mybundle-installer:v1.0`
* `myuser/mybundle@sha256:invocation-image-repo-digest`

That's why I have the continue here on line 434. Apologies for the long tangent if that wasn't what you were asking! 😅

carolynvs: @silvin-lubecki Let me know if that answers your question or if I need to make a change. Would love to merge this PR as it fixes a blocking bug in main. 😁

silvin-lubecki: Ok that makes sense, thanks @carolynvs 👍

silvin-lubecki (approved):
LGTM

PR #233, opened 2020-11-23
Add explicit path to command driver

carolynvs:
Allow a runtime to explicitly set the command driver path, for situations where the driver is not on PATH. Porter is using this to make it easier to run tests in parallel, which doesn't play well with relying on global process values like environment variables.

I refactored the tests to use test cases, and add tests for the new code paths.


vdice (approved):
Nice, this will be a handy addition.

radu-matei (approved):

PR #234, opened 2020-11-23
Write docker output to op writer

### LINKED ISSUE #78: duffle driver ignores operation writer, writes directly to stdout
The docker driver is passed on the `Operation` struct a writer for any output, but instead it is writing directly to stdout. 

As part of Porter we are implementing a verbose flag and would like to be able to show/hide this output, but it is printing directly so we can't control it. Would you be up for a pull request that switches the driver over to using `op.Out`?

carolynvs:
Instead of writing by default to os.Stdout, the docker driver writes to op.Out and the new field op.Err. The op defaults those values to stdout/stderr, so if you weren't setting them previously the behavior is the same though errors are now sent to stderr instead of all lumped into stdout. But for tools like porter that were collecting the output, this enables them to capture all the output and not have rogue lines printed directly to the screen.

This doesn't remove containerOut/containerErr which Docker is relying upon. It should just change the defaults so that the op writers are used first, and are overridden only when SetContainerOut/Err are used.

Closes #78

### DIFF
--- a/action/operation_configs.go
+++ b/action/operation_configs.go
@@ -25,9 +25,12 @@ func (cfgs OperationConfigs) ApplyConfig(op *driver.Operation) error {
 		}
 	}
 
-	// If out wasn't configured, default it
+	// If outputs weren't configured, default them
 	if op.Out == nil {
 		op.Out = os.Stdout
+	}
+	if op.Err == nil {
+		op.Err = os.Stderr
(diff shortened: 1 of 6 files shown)


vdice (approved):
LGTM!

radu-matei (approved):
LGTM

PR #236, opened 2021-01-21
Support in-cluster k8s

carolynvs:
* Add IN_CLUSTER setting to the kubernetes driver.
  The kubernetes driver defaults KUBECONFIG to ~/.kube/config when not set. This prevents the incluster configuration from being used.

  To preserve the existing behavior, I have added IN_CLUSTER which indicates that ambient kubernetes configuration should be used. This is intended for when the cnab process is running inside a pod.
* Do not panic when setting driver config 
  Use the go, Luke. Return errors instead of throwing a panic. This is a breaking change to the driver API, but we aren't at 1.0 and avoiding panics seems worth it.

I didn't see a way to add a test for in-cluster without a ton of extra test infra that we don't have right now. If we are concerned that incluster config needs a test case, we will need to add a docker registry to our kind cluster, build a custom image with a fake cnab-go cli, push it to the kind registry and then run it as a pod. It didn't seem worth it just for this one test. 🤷‍♀️

### DIFF
--- a/driver/debug/debug.go
+++ b/driver/debug/debug.go
@@ -42,6 +42,7 @@ func (d *Driver) Config() map[string]string {
 }
 
 // SetConfig sets configuration for this driver
-func (d *Driver) SetConfig(settings map[string]string) {
+func (d *Driver) SetConfig(settings map[string]string) error {
 	d.config = settings
+	return nil
 }
--- a/driver/kubernetes/kubernetes.go
+++ b/driver/kubernetes/kubernetes.go
@@ -87,7 +87,7 @@ func (k *Driver) Config() map[string]string {
 }
 
 // SetConfig sets Kubernetes driver configuration.
-func (k *Driver) SetConfig(settings map[string]string) {
+func (k *Driver) SetConfig(settings map[string]string) error {
 	k.setDefaults()
 	k.Namespace = settings["KUBE_NAMESPACE"]
 	k.ServiceAccountName = settings["SERVICE_ACCOUNT"]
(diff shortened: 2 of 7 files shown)


vdice (approved):
LGTM

driver/debug/debug.go
```diff
@@ -42,6 +42,7 @@ func (d *Driver) Config() map[string]string {
 }
 
 // SetConfig sets configuration for this driver
-func (d *Driver) SetConfig(settings map[string]string) {
+func (d *Driver) SetConfig(settings map[string]string) error {
```
radu-matei, line 45: Nice catch!
Do we have some tests to make sure the various drivers properly implement the Driver interface?
(of course, not blocking this PR).

driver/kubernetes/kubernetes.go
```diff
@@ -76,6 +78,7 @@ func (k *Driver) Handles(imagetype string) bool {
 // Config returns the Kubernetes driver configuration options.
 func (k *Driver) Config() map[string]string {
 	return map[string]string{
+		"IN_CLUSTER":      "Connect to the ambient cluster",
```
radu-matei, line 81: nit: what is an "ambient" cluster?

radu-matei (approved):

driver/kubernetes/kubernetes.go
carolynvs: Oops, that's not what I meant to write. It should be "Connect to the cluster using in-cluster environment variables"

PR #237, opened 2021-01-26
Make validating a bundle thread-safe

carolynvs:
Background: 
The jsonschema library we use is not thread-safe: https://github.com/qri-io/jsonschema/issues/80. This prevents callers from validating and working with bundles concurrently.

Current Fix:
I made a fork of the latest version jsonschema and tweaked it so that it doesn't use package variables for everything, and uses a lock when it does access the global schema. I'm not sure if that is a solution that they will merge, still waiting on a reply. We are still using unreleased commits on docker repos so at least there is precedent. 😀 

The other changes you see in the PR are due to changes in the latest version of jsonschema.

I tried to fix this without changing jsonschema and wasn't successful. However I am not sure how we feel about using a fork of jsonschema when the patch may never get merged, or the upstream bug fixed.

We really needed this in Porter, so either way I'm maintaining the fork... 🤷‍♀️

### DIFF
--- a/bundle/definition/validators.go
+++ b/bundle/definition/validators.go
@@ -1,9 +1,11 @@
 package definition
 
 import (
+	"context"
 	"encoding/base64"
 	"fmt"
 
+	"github.com/qri-io/jsonpointer"
 	"github.com/qri-io/jsonschema"
 )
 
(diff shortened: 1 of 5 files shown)


bundle/definition/validators.go
```diff
@@ -1,45 +1,54 @@
...
 // NewContentEncoding allocates a new ContentEncoding validator
-func NewContentEncoding() jsonschema.Validator {
+func NewContentEncoding() jsonschema.Keyword {
 	return new(ContentEncoding)
 }
 
-// Validate implements the Validator interface for ContentEncoding
-// which, as of writing, isn't included by default in the jsonschema library we consume
-func (c ContentEncoding) Validate(propPath string, data interface{}, errs *[]jsonschema.ValError) {
+func (c ContentEncoding) Validate(propPath string, data interface{}, errs *[]jsonschema.KeyError) {}
```
vdice, line 20: Is this the standard approach then with the latest jsonschema lib version?  Have `Validate` be a no-op and place all logic in `ValidateKeyword`?

bundle/definition/validators.go
```diff
@@ -1,45 +1,54 @@
...
 		// Add validation support for other encodings as needed
 		// See https://json-schema.org/latest/json-schema-validation.html#rfc.section.8.3
 		default:
-			jsonschema.AddError(errs, propPath, data, fmt.Sprintf("unsupported or invalid contentEncoding type of %s", c))
+			currentState.AddError(data, fmt.Sprintf("unsupported or invalid contentEncoding type of %s", c))
 		}
 	}
 }
 
+func (c ContentEncoding) Register(uri string, registry *jsonschema.SchemaRegistry) {}
```
vdice, line 38: Looks like this method as well as `Resolve` below can/should be no-ops but needed to implement an interface?  Or, will there be scenarios where we'd need to add logic for these?

vdice (commented):
Looks good, just curious about some of the updates per the newer jsonschema lib...

bundle/definition/validators.go
carolynvs: Probably not! Sorry about that, I followed what they did in the tests to define a custom keyword and I didn't dig further. Looks like there may be more that can be done.

carolynvs: Here was their example for making a custom keyword:

https://github.com/qri-io/jsonschema/blob/master/README.md#custom-keywords

They just implemented ValidateKeyword and left the other functions empty.

I looked at a few implementations of string keywords and most also leave those functions empty. 

https://github.com/qri-io/jsonschema/blob/dbc3af1d666cc034a9ac89f10fba0ad6d5cb6c8e/keywords_string.go#L17-L27

An example of a keyword that does implement Register and resolve is `anyOf` which allows you to nest a subschema inside the keyword. That doesn't apply here though.
 
I don't think that the ContentEncoding keyword needs to implement them but if you see something that we should be doing in those functions, let me know. The doc is a bit scarce so I'm piecing it together from usage.

vdice: Ah, thank you for the references.  Indeed, it looks like what you've done for ContentEncoding should be the right approach.  👍

vdice (approved):

radu-matei:
I assume there are not other JSON schema libraries that are thread-safe?

Do you have any preference to maintaining the fork in your account, or in this organization?
(I'm hoping you wouldn't be the only one responsible for updating it if it were in this organization.)

radu-matei (approved):

carolynvs:
The other library is https://github.com/xeipuuv/gojsonschema. They don't say that they are threadsafe, there are still open issues with people asking about that or reporting errors (without reproduction). I like the idea of trying it out and seeing if it would work for us. Let me give that a try and see if it blows up or not.

carolynvs:
Update: Both libraries need more patches to address threadsafety and also neither is making progress towards accepting any patches. 

Another cnab-go user, @dataoleg, has run into [problems](https://github.com/qri-io/jsonschema/issues/80#issuecomment-850302834) with this bug. Since the "better" fix than this patch isn't forthcoming, I would like to see us merge this PR here so that at least cnab-go users are no longer affected.

PR #238, opened 2021-02-01
K8s driver improvements

### LINKED ISSUE #73: [spec compliance] Implement Outputs in Kubernetes Driver

carolynvs:
I am working on a Porter Operator that can run CNAB bundles using K8s CRDs. As part of that I had to make some changes to the Kubernetes driver to get real (not sample) bundles to work.

* Mount a persistent volume to share data: injected files and outputs, between the driver and the bundle job.
* Apply custom labels in the kubernetes driver
  Allow the caller to specify additional labels to apply to resources created by the driver.
* Fix setting CLEANUP_JOBS
  I had the if statement inverted... I've added a test to make sure it (and other settings) are correct.
* Remove requiredCompletions from k8s driver
  The only value that ever makes sense for requiredCompletions is 1 because we aren't running parallel pods for the bundle job.
*  Improve defaulting for kubernetes driver
    * When LimitCPU or LimitMemory is 0, do not set a limit. We don't set a request size, and requesting 0 is not the same as not having a requested resource quantity. This can lead to increased pod evictions.
    * Clarify that LimitCPU and LimitMemory are also used for the requested resources.
    * When ActiveDeadlineSeconds is 0, do not set a deadline.
    * Split reading the configuration in SetConfig and attempting to connect to the cluster, so that we can unit test the settings logic.
    * Move validation for KUBE_NAMESPACE into SetConfig so that any errors can be reported immediately.
    * Add comments for all of the driver settings

With these changes I am able to run @squillace 's kubeflow bundle! 🎉 

```yaml
apiVersion: porter.sh/v1
kind: Installation
metadata:
  name: kubeflow
spec:
  reference: "ghcr.io/squillace/aks-kubeflow-msi:v0.1.7"
  action: "install"
  credentialSets:
    - aks
```

Closes #73

### DIFF
--- a/driver/kubernetes/kubernetes.go
+++ b/driver/kubernetes/kubernetes.go
@@ -108,7 +108,7 @@ func (k *Driver) SetConfig(settings map[string]string) error {
 	k.JobVolumeName = settings["JOB_VOLUME_NAME"]
 
 	cleanup, err := strconv.ParseBool(settings["CLEANUP_JOBS"])
-	if err != nil {
+	if err == nil {
 		k.SkipCleanup = !cleanup
 	}
 
--- a/driver/kubernetes/kubernetes.go
+++ b/driver/kubernetes/kubernetes.go
@@ -3,8 +3,10 @@ package kubernetes
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -12,6 +14,7 @@ import (
 	"time"
 
 	"github.com/docker/distribution/reference"
+	"github.com/hashicorp/go-multierror"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	batchv1 "k8s.io/api/batch/v1"
@@ -49,6 +52,8 @@ type Driver struct {
 	Annotations           map[string]string
 	LimitCPU              resource.Quantity
 	LimitMemory           resource.Quantity
+	JobVolumePath         string
+	JobVolumeName         string
 	Tolerations           []v1.Toleration
 	ActiveDeadlineSeconds int64
 	BackoffLimit          int32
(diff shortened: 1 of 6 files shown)


driver/kubernetes/kubernetes.go
```diff
@@ -44,21 +55,77 @@ var (
...
+
+	// Labels that should be applied to any Kubernetes resources created
+	// by the driver.
+	Labels []string
+
+	// LimitCPU is the amount of CPU to request and the limit for the bundle's job.
+	// Set to zero to not use a limit. Defaults to zero.
+	LimitCPU resource.Quantity
+
+	// LimitMemory is amount of memory to request and the limit for the bundle's job.
```
vdice, line 78:
```suggestion
	// LimitMemory is the amount of memory to request and the limit for the bundle's job.
```

driver/kubernetes/kubernetes.go
```diff
@@ -258,24 +400,109 @@ func (k *Driver) Run(op *driver.Operation) (driver.OperationResult, error) {
...
+
+	opResult, err := k.fetchOutputs(op)
+	if err != nil {
+		opErr = multierror.Append(opErr, err)
+	}
+
+	return opResult, opErr.ErrorOrNil()
+}
+
+func (k *Driver) initJobVolumes(err error) error {
```
vdice, line 432: I'm curious as to why this function takes an `error` as a parameter?  It doesn't seem like we inspect/use it prior to overriding its value.

vdice (commented):
This looks good to me!

The only item we might double-check:

> Remove requiredCompletions from k8s driver

I understand the reasoning supplied, however, are we sure other consumers of this lib weren't somehow relying on this piece of functionality?

driver/kubernetes/kubernetes.go
carolynvs: Oops, that was a mistake. 🤦‍♀️

carolynvs:
requiredCompletions was not exported, so there wasn't any way to set it to anything other than 1. So we had this configurability in place but there was no way to use it (in addition to it not making any sense to change either).

vdice (approved):

radu-matei (approved):

PR #240, opened 2021-02-17
Support Credential.ApplyTo

carolynvs:
Bundles may set the applyTo field on credentials. When it is set, only require the credential when it applies to the specified action.

Implements https://github.com/cnabio/cnab-spec/pull/407


vdice (approved):
🎉 Love the `Scoped` interface -- glad to finally see alignment with these types.  LGTM!

radu-matei (approved):

PR #242, opened 2021-02-26
Check credential.ApplyTo setting credentials on the operation

carolynvs:
When we go through the credentials and set them on the operation,
creating either environment variables or paths, take into account
credential.ApplyTo and do not error out when the credential is
required but doesn't apply to the current action.

This builds on top of my previous PR that added support for
credentials to use ApplyTo, where I had missed that we have duplicate
logic for validating credentials depending on the context (populating
the operation vs validating a credential set).


carolynvs:
Sorry that I didn't get this right the first time! 😓

vdice (approved):

radu-matei (approved):
LGTM

PR #243, opened 2021-03-02
Optionally save logs as output

carolynvs:
When action.SaveLogs is set, a copy of the invocation image logs are saved as an output named `io.cnab.outputs.invocationImageLogs`. The logs are copied as they are written to a logfile, which after the bundle completes, are saved as an output when the claim result is saved.

CNAB tools control if logs are saved or not, defaulting to false.

This is very useful for "remote" bundle execution, like when it's run by the kubernetes driver. The user can see the logs from a run without having to figure out which pod to log, and its stored with the other installation data.

This PR also fixes a bug in the FileSystemStore where listing outputs was not returning any outputs with a period in the name.


claim/result_test.go
```diff
@@ -70,85 +72,196 @@ func TestResult_Validate(t *testing.T) {
...
-		gotContentDigest, ok := outputs.GetContentDigest("test1")
-		require.False(t, ok, "GetContentDigest should report that it did not find the contentDigest")
-		assert.Empty(t, gotContentDigest, "GetContentDigest should return an empty digest when one isn't found")
+		gotValue, ok := outputs.GetMetadata(outputName, metadataKey)
+		require.False(t, ok, "GetMetadata should report that it did not find the value")
+		assert.Empty(t, gotValue, "GetMetadata should return an empty value when one isn't found")
 	})
 
-	t.Run("output has no digest", func(t *testing.T) {
+	t.Run("output has no meatdata", func(t *testing.T) {
```
vdice, line 136: I feel like this is an easter egg for PR reviewers 🥩  and I almost don't want to change it buuut... `s/meatdata/metadata`?

vdice (approved):
LGTM!

claim/result_test.go
carolynvs: I am pro meatdata

carolynvs:
I found a bug with how the filesytem store lists outputs. You wouldn't have noticed unless you were working with an output with a period in the name. So this PR works but if you are using filesystem store, then there's a bug. I'm fixing now.

carolynvs:
@vdice Can you take one more look? I just fixed that bug in FileSystemStore. Seemed like it should be fixed here since without it, when you list outputs on the filesystem, it was missing the logs.

vdice (approved):

carolynvs:
Oh my goodness, sorry one more fix for this FileSystemStore bug. 🤦‍♀️

carolynvs:
Sorry about that, there were a few changes needed to the test and FileSystemStore so that when you list outputs, you get the original filenames. For example, if I save an output named "users.json", when I list outputs, I should see "users.json" as the output name.

vdice (approved):

radu-matei (approved):

PR #246, opened 2021-04-13
Remove unnecessary validation of docker tag on Invocation Images

griffin:
Our image tagging system includes SCM and CI metadata, which changes frequently, which causes an excess of `bundle.json`s to be created. To prevent this, we wanted to omit the tag to keep the `bundle.json` less volatile. 

However, there seems to be a validation, just on Invocation Images, to ensure that the tag is present. This validation seems to have come from a prototype implementation of `duffle`, but I can't find any comments or tests explaining why it's there [1].

As far as I can tell from the spec, omitting a tag is acceptable and is even present in some of the examples for the dependent images [2]. Additionally, pulling by digest is supported by both Docker [3] and the distribution spec [4].

Happy to make an issue if further discussion is warranted, this change seemed like a no-brainer, but happy to comply with your contributions process [5].

[1] https://github.com/cnabio/cnab-go/commit/aa1d2a6e7dd6481db92fc8f5777a58657c0c93e3
[2] https://github.com/cnabio/cnab-spec/blob/cnab-core-1.1.0/101-bundle-json.md#the-image-map
[3] https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier
[4] https://github.com/opencontainers/distribution-spec/blob/v1.0.0-rc2/spec.md#pulling-manifests
[5] https://github.com/cnabio/cnab-go/blob/main/CONTRIBUTING.md

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -139,7 +139,7 @@ func (img *InvocationImage) DeepCopy() *InvocationImage {
 func (img InvocationImage) Validate() error {
 	switch img.ImageType {
 	case "docker", "oci":
-		return validateDockerish(img.Image)
+		return nil
 	default:
 		return nil
 	}
@@ -278,12 +278,5 @@ func (b Bundle) Validate() error {
 		}
 	}
 
-	return nil
-}
-
-func validateDockerish(s string) error {
-	if !strings.Contains(s, ":") {
-		return errors.New("tag is required")
-	}
 	return nil
 }


vdice (commented):
@griffin Thank you for the PR and helpful notes/links/context.  Indeed, the removed code appears to be a relic not necessarily tied to the spec.  When I look at the `image` field description in the spec, it states:

`The image field MUST give a path-like or URI-like representation of the location of the image. It is REQUIRED. The expectation is that an installer should be able to locate the image (given the image type) without additional information.`

Perhaps the last sentence is the closest to _hinting_ that the image be resolvable entirely by itself (sans `digest` or other fields)... but it's vague-ness doesn't seem to bar an image reference with no tag.

Thoughts @carolynvs @radu-matei  ?

carolynvs:
@griffin In my interpretation of the spec, the invocation image should be resolvable using only the image field. If `docker pull INVOCATION_IMAGE` wouldn't work, then it doesn't comply with the spec. Specifying by either digest or tag should be acceptable. The invocation image isn't required to be tagged.

The "dockerish" function takes a shortcut in determining if it is a valid reference and just checks for a colon. This works for both `myimage@sha256:abc123` and `myimage:v1.0.0`. Can you help me understand if you are trying to omit both the digest **and** tag from the image field?

If you can provide the invocationImages subdocument of bundle.json that you are hoping to use (and is failing) that would be helpful!

e.g. 
```
"invocationImages":[
      {
         "contentDigest":"sha256:aaaaaaa...",
         "image":"example/helloworld@sha256:aaaaaaa",
         "imageType":"docker"
      }
   ],
```

griffin:
> In my interpretation of the spec, the invocation image should be resolvable using only the image field.

Ah, I think this is where some of the confusion was for us. We'll add the digest directly to the image ref. The error message in particular had us down a singular train of thought.

Thanks for you help!

PR #251, opened 2021-05-13
feat(schema.go): add support for params of object type in ConvertValue

vdice:
* Adds support for params of type `object` in ConvertValue

(Original issue from Porter: https://github.com/getporter/porter/issues/1459)

### DIFF
--- a/bundle/definition/schema.go
+++ b/bundle/definition/schema.go
@@ -131,6 +131,12 @@ func (s *Schema) ConvertValue(val string) (interface{}, error) {
 		default:
 			return false, errors.Errorf("%q is not a valid boolean", val)
 		}
+	case "object":
+		var obj interface{}
+		if err := json.Unmarshal([]byte(val), &obj); err != nil {
+			return nil, errors.Wrapf(err, "could not unmarshal value %v into a json object", val)
+		}
+		return obj, nil
 	default:
 		return nil, errors.New("invalid definition")
 	}
--- a/bundle/definition/schema_test.go
+++ b/bundle/definition/schema_test.go
@@ -335,16 +335,12 @@ func TestConvertValue(t *testing.T) {
 	is.Error(err)
 
 	pd.Type = "object"
-	_, err = pd.ConvertValue("nope")
-	is.Error(err)
+	out, err = pd.ConvertValue(`{"object": true}`)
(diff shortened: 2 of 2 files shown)


carolynvs (commented):
Do we have existing tests around the validate call for object properties with json schema definitions? Like the replica count property in the original example, if the input was a string instead of an integer, or it violated the maximum, would bundle.validate fail with a good error?

Also have we confirmed what this parameter ends up looking like when it's finally injected into the bundle at runtime? There is a [related porter issue](https://github.com/getporter/porter/issues/1508) about consuming object parameters. I think it would help if we checked how well object parameters are supported from definition through consumption in the bundle to sus out any further changes needed.

vdice:
> Do we have existing tests around the validate call for object properties with json schema definitions? Like the replica count property in the original example, if the input was a string instead of an integer, or it violated the maximum, would bundle.validate fail with a good error?

This library indeed has existing unit tests in this area: https://github.com/cnabio/cnab-go/blob/main/bundle/definition/validation_test.go  These are a bit more abstract in that they build inline JSON Schemas and test them against provided values (using the `func (s *Schema) Validate(data interface{})...` function.  So, the validation has been in place.  This PR adds the link of allowing parameter values of type object to flow through and then be validated by the provided parameter definition schema.

> Also have we confirmed what this parameter ends up looking like when it's finally injected into the bundle at runtime? There is a related porter issue about consuming object parameters. I think it would help if we checked how well object parameters are supported from definition through consumption in the bundle to sus out any further changes needed.

I haven't dug in deeply, but I have tested these changes with a sample bundle like https://github.com/vdice/porter-bundles/blob/master/params/porter.yaml, which will catch instances when a parameter value is invalid per the inline definition schema, etc.  The `ConvertValue` func updated in this PR does a lot of the runtime magic: converting the param vals that have arrived in string form into their intended/native golang types (integer, boolean, object, etc.)

Not sure I've adequately answered any/all of the above.  Let me know!

carolynvs:
That is exactly the info I was looking for, thanks! I really just wanted to make sure that we had tests in place ✅ and have seen this work in Porter ✅ .

carolynvs (approved):
LGTM

radu-matei (approved):

PR #252, opened 2021-05-13
fix(bundle.go): validate param regardless of if override or default

vdice:
* Fixes a bug wherein we weren't previously validating a parameter if we inherit the default.  Now we validate regardless of where we inherit the parameter from (e.g. an override or a default).


carolynvs (approved):
Oops, I imagine this popped up because someone set an invalid default, eh?

Maybe as a follow-up we can check when we validate the bundle? So that someone realizes it's a bad bundle default when they make it instead of waiting until they run the bundle.

vdice:
> Maybe as a follow-up we can check when we validate the bundle? 

Good call!  Will do.

PR #253, opened 2021-05-14
feat(schema.go): add support for params of type array

vdice:
* Adds support for params of type array in the `ConvertValue` function

Builds on/depends on https://github.com/cnabio/cnab-go/pull/251

### DIFF
--- a/bundle/definition/schema.go
+++ b/bundle/definition/schema.go
@@ -131,10 +131,16 @@ func (s *Schema) ConvertValue(val string) (interface{}, error) {
 		default:
 			return false, errors.Errorf("%q is not a valid boolean", val)
 		}
-	case "array", "object":
-		var obj interface{}
+	case "array":
+		var obj []interface{}
 		if err := json.Unmarshal([]byte(val), &obj); err != nil {
-			return nil, errors.Wrapf(err, "could not unmarshal value %v into a json %s", val, dataType)
+			return nil, errors.Wrapf(err, "could not unmarshal value %v into a json array", val)
+		}
+		return obj, nil
+	case "object":
+		var obj map[string]interface{}
+		if err := json.Unmarshal([]byte(val), &obj); err != nil {
+			return nil, errors.Wrapf(err, "could not unmarshal value %v into a json object", val)
 		}
 		return obj, nil
 	default:
--- a/bundle/definition/schema.go
+++ b/bundle/definition/schema.go
@@ -131,10 +131,10 @@ func (s *Schema) ConvertValue(val string) (interface{}, error) {
 		default:
 			return false, errors.Errorf("%q is not a valid boolean", val)
 		}
-	case "object":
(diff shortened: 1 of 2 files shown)


bundle/definition/schema.go
```diff
@@ -131,10 +131,10 @@ func (s *Schema) ConvertValue(val string) (interface{}, error) {
 		default:
 			return false, errors.Errorf("%q is not a valid boolean", val)
 		}
-	case "object":
+	case "array", "object":
```
carolynvs, line 134: oops! I see now that I should have added back number to this switch. I'll update my other PR

bundle/definition/schema_test.go
```diff
@@ -322,17 +322,15 @@ func TestConvertValue(t *testing.T) {
...
-	is.Error(err)
-
-	_, err = pd.ConvertValue("true")
-	is.Error(err)
+	out, err = pd.ConvertValue(`["chocolate", "chip", "cookies"]`)
+	is.NoError(err)
+	is.Equal([]interface{}{"chocolate", "chip", "cookies"}, out)
 
-	_, err = pd.ConvertValue("123.5")
+	out, err = pd.ConvertValue(`["chocolate" "chip" "cookies"]`)
```
carolynvs, line 329: Looks like you removed the other testcases, for "nope", "123", etc. Was that intentional?

vdice: Oh, yes, it was... I did the same when adding support for `object`.  I'd be down to add 'em back for both, though.  WDYT?

carolynvs: In other places we don't as exhaustively check negatives like that, but I don't see the harm in keeping them either? I'll leave it up to you.

radu-matei: Are single values coerced into arrays according to the schema?

vdice: Thanks @carolynvs and @radu-matei .  I've updated the cases for array and object to unmarshal into `[]interface{}` and `map[string]interface{}` respectively, as indeed, if just unmarshal-ing into `interface{}`, there  would be no error if attempting to use `"cookies"` as an array-type parameter.

Let me know how this is looking to y'all now.

bundle/definition/schema.go
```diff
@@ -131,8 +131,14 @@ func (s *Schema) ConvertValue(val string) (interface{}, error) {
...
 		}
+	case "array":
+		var obj []interface{}
+		if err := json.Unmarshal([]byte(val), &obj); err != nil {
+			return nil, errors.Wrapf(err, "could not unmarshal value %v into a json array", val)
+		}
+		return obj, nil
 	case "object":
-		var obj interface{}
+		var obj map[string]interface{}
```
carolynvs, line 141: 👍

carolynvs (approved):
LGTM

PR #254, opened 2021-05-17
feat(*): validate parameter default against its definition

vdice:
* Adds validation against default parameter values when defined

This can be used to catch validation errors at build-time, rather than runtime (which https://github.com/cnabio/cnab-go/pull/252 adds)

### DIFF
--- a/bundle/parameters.go
+++ b/bundle/parameters.go
@@ -1,6 +1,11 @@
 package bundle
 
-import "errors"
+import (
+	"fmt"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+)
 
 // Parameter defines a single parameter for a CNAB bundle
 type Parameter struct {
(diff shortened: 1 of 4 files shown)


bundle/parameters.go
```diff
@@ -23,10 +28,30 @@ func (p *Parameter) AppliesTo(action string) bool {
...
+	// Validate default against definition schema, if exists
+	schema, ok := bun.Definitions[p.Definition]
+	if !ok {
+		return fmt.Errorf("unable to find definition for %s", name)
+	}
+	if schema.Default != nil {
+		var valResult *multierror.Error
+		valErrs, err := schema.Validate(schema.Default)
+		if err != nil {
+			valResult = multierror.Append(valResult, errors.Wrapf(err, "encountered an error validating parameter %s", name))
```
carolynvs, line 45: Let's sneak something about this being a default value for the parameter being validated. So that it's clear exactly which part of the parameter is messed up.

```suggestion
			valResult = multierror.Append(valResult, errors.Wrapf(err, "encountered an error validating the default value %v for parameter %s", schema.Default, name))
```

bundle/parameters.go
```diff
@@ -23,10 +28,30 @@ func (p *Parameter) AppliesTo(action string) bool {
...
+		}
+		for _, valErr := range valErrs {
+			valResult = multierror.Append(valResult, fmt.Errorf("encountered validation error for parameter %s: %v", name, valErr.Error))
```
carolynvs, line 48:
```suggestion
			valResult = multierror.Append(valResult, fmt.Errorf("encountered validation error for the default value %v for parameter %s: %v", schema.Default, name, valErr.Error))
```

bundle/parameters.go
```diff
@@ -23,10 +28,30 @@ func (p *Parameter) AppliesTo(action string) bool {
...
+		}
+		if valResult != nil {
```
carolynvs, line 50: I think a safer way to check this, so that doesn't return when the multierror is initialized but empty would be:

```suggestion
		if valResult.ErrorOrNil() != nil {
```

bundle/parameters.go
vdice: Oops, haha, had thought of this and pushed a commit without checking in on this review.  Good call.  I'll incorporate the full suggestion you've added.

carolynvs (approved):
LGTM

PR #255, opened 2021-05-17
feat(outputs.go): add output validation

vdice:
* Adds validation of a bundle's outputs, including validating that the default value (if provided) adheres to the output schema


carolynvs (approved):
LGTM!

PR #256, opened 2021-05-17
feat(validation.go): add ValidateSchema function

vdice:
This adds a `ValidateSchema` method to the `Schema` struct.  This will be handy for validating that the schema itself is well-formed and valid JSON Schema.

Essentially, we were already doing this schema validation, but only in the context of a value we wanted to validate against the schema.  Therefore, I wanted to break it apart for standalone use.  For instance, to validate a parameter definition at build time and when we have no value (default or otherwise) to validate with.

As follow-ups, I hope to utilize this new method when validating that the definition schemas for parameters and outputs in a bundle are valid JSON Schema.  (From there, we can actually validate that any supplied parameter or output defaults are valid per these validated schemas -- we'll have added these pieces in PRs https://github.com/cnabio/cnab-go/pull/254 and https://github.com/cnabio/cnab-go/pull/255)

### DIFF
--- a/bundle/definition/validation.go
+++ b/bundle/definition/validation.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	"github.com/pkg/errors"
+	"github.com/qri-io/jsonschema"
 )
 
 // ValidationError error represents a validation error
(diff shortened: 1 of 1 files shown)


carolynvs (approved):
LGTM

PR #269, opened 2021-08-26
Reference images by digest in docker driver

HadrienPatte:
This is a new take on #166 with the same goal: reference images by digest in the docker driver, similarly to how it's currently done in the kubernetes driver.


carolynvs:
@HadrienPatte I was just checking in to see if the workaround I suggested makes sense? The CNAB General Meeting is tomorrow (https://cnab.io/community-meetings/) and we could discuss this more at the meeting if that would be helpful, especially if we want to try to clarify the spec.

HadrienPatte:
Hi, thank you for suggesting this solution, I changed the way we generate `bundle.json` files to have a fully digested ref in the `invocationImage.[0].image` field, and with this we don't need the patch that this PR brings anymore.
I still think that using the provided `contentDigest` for more than just validation, and to enforce pulling and running an image with that digest would be a useful feature. If this requires a spec change, I'm happy to discuss this more with the CNAB community.

carolynvs:
What is nice about your suggested change to the spec is that it prevents a bundle from getting into an "unrunnable state" which kind of defeats some of the purpose of making a bundle in the first place.

1. publish a bundle where the image reference uses a tag.
2. people use the bundle and it works. yay!
3. the developer makes a mistake and force pushes over the invocation image.
4. the same people who previously could run the bundle will now have the bundle fail because cnab-go will pull using the image tag, and then fail the digest comparison.

If we changed the spec to have images (both invocation and referenced images) always be by digest, then you couldn't get into this situation.

Next step would be to open an issue in https://github.com/cnabio/cnab-spec and propose your change to let others weigh in on it.

HadrienPatte:
The proposed workaround works well for our use case, so I'm gonna close this PR as we don't need it anymore, thanks for the reviews :)

PR #272, opened 2021-11-10
Add bundle.GetAction helper

carolynvs:
Add GetAction function to bundle that helps retrieve the definition of an action, including core actions: install, upgrade, uninstall. This allows the client to use common logic for some checks about what should happen during an action that is stateless or modifies bundle resources.

For example, should a claim be recorded for a particular action or not.


carolynvs:
The build will be fixed by #273. So that should merge first.

bundle/bundle_test.go
```diff
@@ -949,3 +949,40 @@ func TestBundle_IsOutputSensitive(t *testing.T) {
...
+
+func TestBundle_GetAction(t *testing.T) {
+	testcases := []struct {
+		action    string
+		modifies  bool
+		stateless bool
+		wantErr   string
+	}{
+		{action: "install", modifies: true},
+		{action: "install", modifies: true},
```
vdice, line 961: Perhaps cases 2 and 3 were meant to have `action` as "upgrade" and "uninstall" ?

carolynvs: I just really wanted to test install like _extra_ hard... 🤦‍♀️

carolynvs: Fixed!

vdice (approved):

PR #274, opened 2021-11-10
Change DefaultSchemaVersion to not return an error

carolynvs:
There isn't really a need to force people to handle (or ignore) an error from DefaultSchemaVersion. The default schema is hard coded and with tests and buildtime asserts, we know that it was set to a valid value.

We were already throwing a panic or ignoring the error anyway at runtime, so this is more of an assurance actually than what we had before.


vdice (approved):
👍 Great improvement.

PR #276, opened 2021-11-10
Remove crud data access layer

### LINKED ISSUE #275: Remove CRUD data access layer
When we originally created this library, it was a lift and shift of code from duffle. This ended up bringing over things that are not related to the cnab spec or runtime, such as the crud data access layer. Then as we added more complex features to the runtime, I updated the action struct to interact with the CRUD layer, to help with common tasks like storing claims and results from a run.

The thing is that it never really belonged in this library to begin with. Every client has different needs and we ended up with a lowest common denominator that isn't useful to anyone (and is slow too). DataDog never used the data access layer, and Porter had to scrap it months ago.

At this point its a bit of a vestigial tail and I'd like to get rid of it. 😀 

Is anyone using it and would care if it was removed?

carolynvs:
* Remove any data providers/stores and the general utility CRUD stores.
* Update examples to demonstrate how to use cnab-go when you are
responsible for persisting the claim yourself.

Closes #275


vdice (approved):
We've yet to seen any feedback on https://github.com/cnabio/cnab-go/issues/275 pushing to keep this logic; LGTM.

radu-matei (approved):

jlegrone (approved):

PR #281, opened 2022-03-14
Update go.mod

carolynvs:
This updates our go.mod to use recent security patches for Go, docker, containerd, and kubernetes. Also now that we are on a newer version of Go, I've replaced the packr library with go:embed.


carolynvs:
Moving back to draft as there's something about this version bump that seriously upsets go mod tidy when I use this change in Porter. 🤔

vdice (approved):

PR #282, opened 2022-04-18
secrets/host: implement Create method for host plugin

VinozzZ:
Implement a Create method so it matches with the new secret plugin interface in porter

### DIFF
--- a/secrets/host/host.go
+++ b/secrets/host/host.go
@@ -1,6 +1,7 @@
 package host
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


carolynvs:
Since the create method is just for Porter, and the rest of cnab-go wouldn't take advantage of it, let's see if we can make this change in just porter alone, and not edit cnab-go.

In Porter, we can create a new type that is composed of `SecretStore` and add the Create method on the wrapping type. Then Porter can use that wrapped type instead of SecreStore directly. Can you give that a try?

VinozzZ:
Thanks for the suggestion. I will close this PR since this is only useful for porter

PR #284, opened 2022-06-03
Support connecting to TLS secured docker host

carolynvs:
When we create a docker cli instance using just the docker/cli library, some important configuration doesn't happen, namely reading DOCKER_TLS_VERIFY and DOCKER_CERT_PATH. Unlike DOCKER_HOST, these other variables for connecting to a TLS secured docker host are not configured in the main library functions but instead just in the CLI's main package when the flags (for stuff like --tlsverify) are bound.

This means we need to configure this ourselves when using that library. I've added a function that consolidates all the necessary configuration steps to make a docker cli client that is configured the same as if running the docker CLI directly on your computer.

I've tested this manually against a remote docker host secured with a self-signed certificate and I'm able to build, push and run bundles on a the remote host with this fix.


carolynvs:
cc @vinozzz This is the patch that I'm relying on for https://github.com/getporter/porter/pull/2134

carolynvs:
Okay, rebased now that the build is fixed and this is ready for review.

vdice (approved):

PR #289, opened 2022-06-10
Add github action that builds project

carolynvs:
I've converted our Azure DevOps Pipeline to a GitHub Workflow. I split out calling each target so that it's easier to find which step failed and repeat just failing steps in the build.

This is using the newer setup-go github action that has built in support for detecting which version of go to use based on the go.mod file and also caching downloaded go mods to speed up subsequent builds.

I've found that the github actions build much more quickly than pipelines and are easier for non-MSFT people to manage. Due to a quirk in how DevOps authenticates users, it was only this week that I was able to get my personal account to be able to log into DevOps to do things like re-try failed builds.

What do you think about adding this, seeing how it compares to our current pipeline (i.e. they both fail/pass for the same commits correctly), and then switching over completely?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,28 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: build
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


vdice (approved):
I'm definitely in support of at least comparing them and do ultimately favor GH Actions for this repo.  LGTM.

PR #290, opened 2022-06-17
fix: avoid erasing local schema if fetching schema from cdn.cnab.io fails

### LINKED ISSUE #287: Flake: TestClaimSchema
Lately I'm seeing TestClaimSchema fail just on CI. When I run it locally, it always passes (and has passed just fine previously).

Our dependency for jsonschema hasn't changed in a long time so I'm not sure what's causing the EOF.

```
=== RUN   TestClaimSchema
    claim_test.go:224: 
        	Error Trace:	claim_test.go:224
        	Error:      	Received unexpected error:
        	            	EOF
        	            	unable to compile schema validator for schema
        	            	
        	            	github.com/cnabio/cnab-go/schema.Validate
        	            		/home/vsts/work/1/s/schema/validation.go:50
        	            	github.com/cnabio/cnab-go/schema.ValidateClaim
        	            		/home/vsts/work/1/s/schema/validation.go:30
        	            	github.com/cnabio/cnab-go/claim.TestClaimSchema
        	            		/home/vsts/work/1/s/claim/claim_test.go:223
        	            	testing.tRunner
        	            		/opt/hostedtoolcache/go/1.17.8/x64/src/testing/testing.go:1259
        	            	runtime.goexit
        	            		/opt/hostedtoolcache/go/1.17.8/x64/src/runtime/asm_amd64.s:1581
        	Test:       	TestClaimSchema
        	Messages:   	failed to validate claim schema
--- FAIL: TestClaimSchema (0.00s)

```

epsxy:
# Motivation

Fixes #287

Everything is detailled in [this comment](https://github.com/cnabio/cnab-go/issues/287#issuecomment-1158289217), but, TLDR: The CI is using the `fetch-schemas` go program to download the latest schema data files from the cnab cdn. If it doesn't reply for some reason, the program is erasing the local version of the schema file by an empty file, which can cause 2 tests to fail:

- `TestClaimSchema`
- `TestBundleSchema`

This PR is adding a return when the fetch request fails, so that we don't write an empty file on the disk when it fails and so that the tests don't fail in that case.

## Reproducing the issue

The issue can be reproduced by running locally `go run fetch-schemas.go`. If you run this command before this PR, after turning off your internet connection (eg wifi), you will see that both files gets erased. After the PR, the files are not updated anymore (with an empty content) when you run this command without an internet connection.

## Discussion

There's still a possibility that we don't want that PR, especially in the case that we want to be conservative and fail if we don't have the exact version of the spec we need (especially if we don't think the version in the repo is up to date). In that case maybe we would like the tests to fail rather than maybe releasing a false negative (if the version in the repo is not the same than the version upstream on the CDN).

### DIFF
--- a/fetch-schemas.go
+++ b/fetch-schemas.go
@@ -27,6 +27,8 @@ func main() {
 		bytes, err := fetchSchema(schema, version)
 		if err != nil {
 			fmt.Printf("unable to fetch %s schema with version %s: %s\n", schema, version, err.Error())
+			// if cdn.cnab.io is not reachable, we stick with default schema
+			return
 		}
 
 		err = writeSchema(schema, bytes)
--- a/fetch-schemas.go
+++ b/fetch-schemas.go
@@ -28,7 +28,7 @@ func main() {
 		if err != nil {
 			fmt.Printf("unable to fetch %s schema with version %s: %s\n", schema, version, err.Error())
 			// if cdn.cnab.io is not reachable, we stick with default schema
-			return
+			continue
 		}
 
 		err = writeSchema(schema, bytes)


carolynvs (approved):
Thanks for the extensive research and explanation of the root cause!

PR #306, opened 2024-10-15
Bumped all deps with an RHDA warning

dgannon991:
Quite a few of the dependencies where flagging up as having critical vulnerabilities. This was causing downstream projects to also flag for the same. I've just bumped the failing dependencies and made some tiny changes to support these new libraries (nearly all were just methods or types moving to new places)


urfinjuezz (approved):

schristoff (approved):
build is passing, golint is mad but was fixed in a different PR

